PR #2, opened 2017-01-12
Implement Store, BaseQuery and ResultSet

erizocosmico:
`model` and `id` and `schema` are not final implementations. I needed the actual implementation of them in order for the rest of stuff to work, but the final implementation will be the one from @dpordomingo


store.go
```diff
@@ -1,199 +1,207 @@
...
 }
 
-// Insert insert the given document in the collection, returns error if no-new
-// document is given. The document id is setted if is empty.
-func (s *Store) Insert(doc DocumentBase) error {
-	if !doc.IsNew() {
+// Insert insert the given record in the table, returns error if no-new
+// record is given. The record id is set if it's empty.
+func (s *Store) Insert(record Record) error {
+	if !record.IsPersisted() {
```
dpordomingo: isNew != isPersisted
then... it should be 
`if record.IsPersisted() {
    return ErrNonNewDocument
}`

store.go
```diff
@@ -1,199 +1,207 @@
...
-func (s *Store) Save(doc DocumentBase) (updated bool, err error) {
-	id := doc.GetId()
-	if len(id) == 0 {
-		return false, ErrEmptyID
+// Update updates the given fields of a record in the table. All fields are
+// updated if no fields are provided. For an update to take place, the record is
+// required to have a non-empty ID and not to be a new record.
+// Returns the number of updated rows and an error, if any.
+func (s *Store) Update(record Record, cols ...string) (int64, error) {
+	if record.IsPersisted() {
```
dpordomingo: isNew != isPersisted
then... it should be 
`if !record.IsPersisted() {
    return 0, ErrNewDocument
}`

store.go
```diff
@@ -1,199 +1,207 @@
...
+		Exec()
+	if err != nil {
+		return err
 	}
 
-	sess, c := s.getSessionAndCollection()
-	defer sess.Close()
-
-	return c.UpdateId(doc.GetId(), doc)
+	record.setPersisted(false)
```
dpordomingo: isNew != isPersisted
then... it should be 
`record.setPersisted(true)`

store.go
```diff
@@ -1,199 +1,207 @@
...
+// Save inserts or updates the given record in the table. It requires a record
+// with non empty ID.
+func (s *Store) Save(record Record) (updated bool, err error) {
+	if record.GetID().IsEmpty() {
+		return false, ErrEmptyID
 	}
 
-	if q.GetLimit() != 0 {
-		mq.Limit(q.GetLimit())
+	if record.IsPersisted() {
```
dpordomingo: isNew != isPersisted
then... it should be 
`if !record.IsPersisted() {return false, s.Insert(record)}`

erizocosmico:
Fixed @dpordomingo

model.go
```diff
@@ -0,0 +1,89 @@
+package kallax
+
+import (
+	"database/sql"
+	"database/sql/driver"
+
+	uuid "github.com/satori/go.uuid"
+)
+
+type ID uuid.UUID
```
Serabe: Is this the _de facto_ package for UUID?

model.go
```diff
@@ -0,0 +1,89 @@
...
+func (id ID) IsEmpty() bool {
+	return uuid.UUID(id) == uuid.Nil
+}
+
+type Model struct {
+	ID        ID
+	persisted bool
+}
+
+func (m *Model) GetID() ID {
```
Serabe: Isn't the idiom for getters to just be the name of the property?

model.go
```diff
@@ -0,0 +1,89 @@
...
+func (m *Model) SetID(id ID) {
+	m.ID = id
+}
+
+func (m *Model) IsPersisted() bool {
+	return m.persisted
+}
+
+func (m *Model) setPersisted(isPersisted bool) {
+	m.persisted = isPersisted
```
Serabe: I would say that something that is persisted cannot be set as not persisted, so maybe just `setPersisted` without params.

operators.go
```diff
@@ -0,0 +1,118 @@
...
+func And(conds ...Condition) Condition {
+	return and{squirrel.And(condsToSqlizers(conds))}
+}
+
+// Not returns the given condition negated.
+func Not(cond Condition) Condition {
+	return not{cond}
+}
+
+// Lt returns a condition that will be true when `col` is lower than `value`.
```
Serabe: Wrong documentation.

query.go
```diff
@@ -1,106 +1,203 @@
...
+			Select().
+			From(table),
+		batchSize: 50,
 	}
+}
 
-	return operators.And(q.clauses...)
+// Select adds the given columns to the list of selected columns in the query.
+func (q *BaseQuery) Select(columns ...string) {
+	q.columns.add(columns...)
```
Serabe: Shouldn't `Select` remove the columns from the excluded set as well?

Think of:

```go
query.SelectNot('name')
// In other place
query.Select('name')
```

User should be able to select columns even if those were excluded before.

query.go
```diff
@@ -1,106 +1,203 @@
...
 }
 
-// Sort sets the sorting cristeria of the query.
-func (q *BaseQuery) Sort(s Sort) {
-	q.sort = s
+// SelectNot adds the given columns to the list of excluded columns in the query.
+func (q *BaseQuery) SelectNot(columns ...string) {
+	q.excludedColumns.add(columns...)
```
Serabe: See comment before. Though the usual processing steps would include a `select - excluded` algorithm, that would tie two methods, adding an implicit dependency between them.

query.go
```diff
@@ -1,106 +1,203 @@
...
+	q.builder = q.builder.Where(squirrel.Sqlizer(cond))
+}
+
+// compile returns the selected column names and the select builder.
+func (q *BaseQuery) compile() ([]string, squirrel.SelectBuilder) {
+	columns := q.selectedColumns()
+	return columns, q.builder.Columns(columns...)
+}
+
+// String returns the SQL generated by the
```
Serabe: `by the`...

query_test.go
```diff
@@ -1,27 +1,77 @@
...
+
+	s.Equal(s.q, copy)
+	s.NotEqual(unsafe.Pointer(s.q), unsafe.Pointer(copy))
+}
+
+func (s *QuerySuite) TestSelectedColumns() {
+	s.q.Select("a", "b", "c")
+	s.q.SelectNot("b")
+	s.Equal([]string{"a", "c"}, s.q.selectedColumns())
+}
```
Serabe: Add tests for `Select + SelectNot + Select` where at least one of the fields is passed to the three calls. Same for `SelectNot + Select + SelectNot`.

resultset.go
```diff
@@ -1,87 +1,54 @@
...
-	//ErrNotFound document not found
-	ErrNotFound = errors.New("document not found")
-	// ErrStop if is used on a callback of a ResultSet.ForEach function the loop
-	// is stopped
-	ErrStop = errors.New("document not found")
-)
+// ErrRawScan is an error returned when a the `Scan` method of `ResultSet`
+// is called with a `ResultSet` created as a result of a `RawQuery`, which is
+// not allowed.
+var ErrRawScan = fmt.Errorf("result set comes from raw query, use RawScan instead")
```
Serabe: Shouldn't we use Máximo's package?

store.go
```diff
@@ -1,199 +1,207 @@
...
+	result, err := s.updater.
+		SetMap(clauses).
+		Where(squirrel.Eq{s.schema.Identifier(): record.GetID()}).
+		Exec()
+	if err != nil {
+		return 0, err
+	}
 
-	return c.RemoveId(doc.GetId())
+	return result.RowsAffected()
```
Serabe: Can this be other than 0 or 1? If it is 0, should'n that be an error?

model.go
dpordomingo: I found it as the best uuid package too

model.go
dpordomingo: using ID() would conflict with the homonym ID property

model.go
dpordomingo: I did it in my branch, that I'll pullrequest during the day :)

model.go
erizocosmico: yep

model.go
erizocosmico: yep, this is not a final implementation of model

query.go
erizocosmico: totally right

query.go
erizocosmico: lol

resultset.go
erizocosmico: is it ready yet?

store.go
erizocosmico: Yep, maybe it should

erizocosmico:
@Serabe we'll add the testing thingy later to the README along with all documentation

erizocosmico:
Fixed @Serabe

resultset.go
Serabe: Not sure :$

PR #3, opened 2017-01-16
Add wrapper SQL types

erizocosmico:
Only last commit is from this PR

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -1,0 +1,166 @@
+package types
+
+import (
+	"database/sql"
+	"database/sql/driver"
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"reflect"
+
+	"github.com/lib/pq"
+)
+
+// URL is a wrapper of url.URL that implements SQLType interface.
+type URL url.URL
+
+func (u *URL) Scan(v interface{}) error {
+	switch t := v.(type) {
+	case []byte:
+		return u.Scan(string(t))
+	case string:
+		url, err := url.Parse(t)
+		if err != nil {
+			return fmt.Errorf("error scanning url: %s", err)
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -0,0 +1,227 @@
...
+	bytes, err := json.Marshal(v)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
+}
+
+// SQLType is the common interface a type has to fulfill to be considered a
+// SQL type.
+type SQLType interface {
```
erizocosmico: If find `SQLType` way more explicit than `ScanValuer`

types/types.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return fmt.Errorf("cannot scan type %s into JSON type", reflect.TypeOf(v))
+}
+
+// JSONValue converts something into json.
+// WARNING: This is here temporarily, might be removed in the future, use
+// `JSON` instead.
+func JSONValue(v interface{}) (driver.Value, error) {
+	bytes, err := json.Marshal(v)
```
erizocosmico: right

erizocosmico:
Done @dpordomingo

types/types.go
```diff
@@ -0,0 +1,166 @@
...
+	case string:
+		url, err := url.Parse(t)
+		if err != nil {
+			return fmt.Errorf("error scanning url: %s", err)
+		}
+
+		*u = URL(*url)
+		return nil
+	}
+	return fmt.Errorf("cannot scan type %s into URL type", reflect.TypeOf(v))
```
Serabe: What is the difference between `reflect.TypeOf` and using `%T`?

PR #5, opened 2017-01-17
Add Hooks

dpordomingo:


behaviours/common.go
```diff
@@ -0,0 +1,72 @@
+package behaviours
+
```
erizocosmico: move all the package to the root, we don't want to go importing a bazillion packages for very core functionality

behaviours/identifiable.go
```diff
@@ -0,0 +1,45 @@
+package behaviours
+
+import "github.com/src-d/go-kallax/common"
+
+//Identifiable must be implemented by those values that can be identified by an ID
+type Identifiable interface {
+	// GetID returns the ID.
+	GetID() common.ID
```
erizocosmico: `ID` should be in root, not in `common` package

behaviours/identifiable.go
```diff
@@ -0,0 +1,45 @@
...
+	// SetID overrides the ID.
+	SetID(id common.ID)
+	// Identify sets the ID if it does not exist and returns the ID of the object
+	Identify() common.ID
```
erizocosmico: why do you need this?

behaviours/identifiable.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Identificator modelates an object that knows about its ID and generates it if neccessary
+type Identificator struct {
```
erizocosmico: why do you need this?

behaviours/timestampable.go
```diff
@@ -0,0 +1,37 @@
...
+
+import "time"
+
+// Timestampable must be implemented by those object that can be timestamped
+type Timestampable interface {
+	Timestamp()
+}
+
+// Timestamp modelates an object that knows about when was created and updated
+type TimestampDates struct {
```
erizocosmico: i'd name this `Timestamps` or `Timestampable`

behaviours/timestampable.go
```diff
@@ -0,0 +1,37 @@
+package behaviours
+
+import "time"
+
+// Timestampable must be implemented by those object that can be timestamped
+type Timestampable interface {
```
erizocosmico: why do you need this interface?

common/id.go
```diff
@@ -0,0 +1,30 @@
+package common
```
erizocosmico: move to root

model.go
```diff
@@ -1,63 +1,38 @@
...
-
-func (id ID) IsEmpty() bool {
-	return uuid.UUID(id) == uuid.Nil
-}
-
+// Model is the base type of the items that are stored
 type Model struct {
-	ID        ID
+	behaviours.Identificator
+	behaviours.TimestampDates
```
erizocosmico: timestampable should not be included by default

model.go
```diff
@@ -1,63 +1,38 @@
...
-}
-
-func (id ID) IsEmpty() bool {
-	return uuid.UUID(id) == uuid.Nil
-}
-
+// Model is the base type of the items that are stored
 type Model struct {
-	ID        ID
+	behaviours.Identificator
```
erizocosmico: put `ID` field here, it's not necessary to embed anything here

model.go
```diff
@@ -1,63 +1,38 @@
...
-}
-
+// IsPersisted returns whether this Model is new in the store or not.
 func (m *Model) IsPersisted() bool {
 	return m.persisted
 }
 
-func (m *Model) setPersisted(isPersisted bool) {
-	m.persisted = isPersisted
+func (m *Model) setPersisted() {
```
erizocosmico: if `setPersisted` is not really a `set` method I would name it `persist` or `persisted`

model.go
```diff
@@ -66,11 +41,33 @@ func (m *Model) setWritable(w bool) {
...
+type Persistable interface {
+	// IsPersisted returns whether this Model is new in the store or not.
+	IsPersisted() bool
+	setPersisted()
+}
+
+// Writable must be implemented by those values that defines internally
+//  if they can be sent back to the database to be stored with its changes.
+type Writable interface {
+	IsWritable() bool
```
erizocosmico: should have comment

store.go
```diff
@@ -53,9 +53,7 @@ func (s *Store) Insert(record Record) error {
 		return ErrNonNewDocument
 	}
 
-	if record.GetID().IsEmpty() {
-		record.SetID(NewID())
-	}
+	record.Identify()
```
erizocosmico: I would keep it as it was. It's not going to be used anywhere else and feels a little pointless to keep this method just for this ad-hoc logic

behaviours/common.go
```diff
@@ -0,0 +1,72 @@
...
+// Hookable must be implemented by those entities that attend to all DB Hooks
```
erizocosmico: why do you need the hookable interface? The thing about events is that you should be able to implement just the ones you need. This interface forces you to implement all of them

behaviours/common.go
```diff
@@ -0,0 +1,72 @@
...
+	BeforeUpdater
+	AfterUpdater
+	BeforeInserter
+	AfterInserter
+	BeforeDeleter
+	AfterDeleter
+}
+
+// AfterLoader must be implemented by those entities listening the AfterLoad event
+type AfterLoader interface {
```
erizocosmico: i don't know if we are going to need these interfaces at all. The code will be generated and we will not check if a record fulfills them, so the only use is if someone maybe wants to check it by hand

model.go
Serabe: `persisted` does not sound ok for a method that does something

dpordomingo: `person.Persist()` should save the person into the db, what is not the purpose of this method.

dpordomingo:
@erizocosmico the proposed changes was applied

model.go
erizocosmico: ok, fair enough :+1:

model.go
```diff
@@ -41,6 +42,25 @@ func (m *Model) setWritable(w bool) {
 	m.writable = w
 }
 
+// GetID returns the ID.
+func (i *Model) GetID() ID {
+	return i.ID
+}
+
+// SetID overrides the ID.
+//  The ID should not be modified once it has been set and stored in the DB
```
erizocosmico: extra space before the "The"

model.go
```diff
@@ -41,6 +42,25 @@ func (m *Model) setWritable(w bool) {
...
+func (i *Model) SetID(id ID) {
+	i.ID = id
+}
+
+//Identifiable must be implemented by those values that can be identified by an ID
```
erizocosmico: missing space after `//`

model.go
```diff
@@ -51,24 +71,26 @@ type Persistable interface {
...
 	setWritable(bool)
 }
 
 // ColumnAddresser must be implemented by those values that exposes its properties
 //  under pointers, identified by its property names
 type ColumnAddresser interface {
-	// ColumnAddress returns a pointer to the object property identified by the passed string
-	//  or an error if that property does not exist
+	// ColumnAddress returns a pointer to the object property identified by the
+	// passed string or an error if that property does not exist
```
erizocosmico: i'd say "column name" instead of "passed string"

resultset.go
```diff
@@ -10,6 +10,10 @@ import (
 // not allowed.
 var ErrRawScan = fmt.Errorf("result set comes from raw query, use RawScan instead")
 
+// ErrStop is an error returned by a callback to stop the iteration in a
```
erizocosmico: you can remove this one from here, I added it somewhere else on the codegen branch (or I can delete mine and move this one to `store.go`)

model.go
Serabe: `setAsPersisted`? Wasn't we naming that method `Save`?

resultset.go
dpordomingo: Removed

PR #8, opened 2017-01-19
Implement code generation of models

erizocosmico:


generator/processor.go
```diff
@@ -110,240 +102,293 @@ func (p *Processor) parseSourceFiles(filenames []string) (*types.Package, error)
...
 
-	for _, code := range p.SourceCode {
-		if re.Match(code) {
+		if strings.HasPrefix(string(e), "Before") {
+			if signature.Results().Len() != 1 ||
+				!isBuiltinError(signature.Results().At(0).Type()) {
+				return false
+			}
 			return true
 		}
```
Serabe: Outdated!

generator/types.go
```diff
@@ -9,90 +10,173 @@ import (
...
-	"struct":                        true,
-	"bool":                          true,
-	"map":                           true,
-	"time.Time":                     true,
-	"interface{}":                   true,
-	"gopkg.in/mgo.v2/bson.ObjectId": true,
-}
-
+// https://www.postgresql.org/docs/current/static/sql-keywords-appendix.html
+var reservedKeywords = map[string]struct{}{
```
Serabe: Sort alphabetically, please

operators.go
```diff
@@ -7,110 +7,97 @@ import (
...
+	return func(schema Schema) squirrel.Sqlizer {
+		return not{cond(schema)}
+	}
 }
 
 // In returns a condition that will be true when `col` is equal to any of the
 // passed `values`.
-func In(col string, values []interface{}) Condition {
-	return eq{squirrel.Eq{col: values}}
+func In(col SchemaField, values ...interface{}) Condition {
```
dpordomingo: we'd preffer `func In(col SchemaField, values []interface{}) Condition {`

operators.go
```diff
@@ -7,110 +7,97 @@ import (
...
+	return func(schema Schema) squirrel.Sqlizer {
+		return squirrel.Eq{col.QualifiedName(schema): values}
+	}
 }
 
 // NotIn returns a condition that will be true when `col` is distinct to all of the
 // passed `values`.
-func NotIn(col string, values []interface{}) Condition {
-	return neq{squirrel.NotEq{col: values}}
+func NotIn(col SchemaField, values ...interface{}) Condition {
```
dpordomingo: func NotIn(col SchemaField, values []interface{}) Condition {

generator/cli/kallax/cmd_generate.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 type CmdGenerate struct {
-	Input  string `short:"" long:"input" description:"input package directory" default:"."`
-	Output string `short:"" long:"output" description:"output file name" default:"kallax.go"`
+	Input  string `short:"i" long:"input" description:"input package directory" default:"."`
```
mcuadros: when an argument has a default usually doesn't have a short version

schema.go
```diff
@@ -1,8 +1,72 @@
...
+
+// Schema represents a table schema in the database. Contains some information
+// like the table name, its columns, its identifier and so on.
 type Schema interface {
-	Alias() string
-	Table() string
-	Identifier() string
-	Columns() []string
+	// GetAlias returns the name of the alias used in queries for this schema.
+	GetAlias() string
```
mcuadros: why we added here `Get` prefix? if you have a implementation of this interface with the name `Alias` make it private, but don't add this ugly `Get` prefix

erizocosmico: Then you'd have to make a constructor with a bazillion arguments instead of just constructing the type.

Also, the type that implements `Schema` is embedded in the generated schema, which can override the method by having an `Alias` field in the model.

```go
type Foo struct {
  kallax.Model
  Alias string
}
```

would generate a schema like

```
type fooSchema struct {
  *kallax.BaseSchema
  Alias kallax.SchemaField
}
```

And the Alias field has preference over the method when calling `schema.Alias()`

The schema methods is not something you use manually, they are only used internally by the generated code, so that's why we preferred to add `Get` before the method name and avoid possible name conflicts.

PR #22, opened 2017-01-19
Implement Reload method on stores. Closes #13

### LINKED ISSUE #13: Store.Reload method

erizocosmico:
Closes #13

### DIFF
--- a/store.go
+++ b/store.go
@@ -211,6 +211,31 @@ func (s *Store) MustFind(q Query) *ResultSet {
 	return rs
 }
 
+// Reload refreshes the record with the data in the database and makes the
+// record writable.
+func (s *Store) Reload(record Record) error {
+	if record.GetID().IsEmpty() {
+		return ErrEmptyID
+	}
+
+	q := NewBaseQuery(s.schema)
+	q.Where(Eq(s.schema.GetID(), record.GetID()))
+	q.Limit(1)
+	columns, builder := q.compile()
+
+	rows, err := builder.RunWith(s.db).Query()
+	if err != nil {
+		return err
+	}
+
+	rs := NewResultSet(rows, false, columns...)
+	if !rs.Next() {
+		return sql.ErrNoRows
+	}
+
+	return rs.Scan(record)
+}
+
 // Count returns the number of rows selected by the given query.
 func (s *Store) Count(q Query) (count int64, err error) {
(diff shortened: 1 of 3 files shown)


store_test.go
```diff
@@ -169,6 +169,27 @@ func (s *StoreSuite) TestCount() {
...
+
+	s.False(model.IsWritable())
+	s.Equal(0, model.Age)
+
+	s.Nil(s.store.Reload(&model))
+
+	s.True(model.IsWritable())
+	s.Equal(1, model.Age)
+}
+
```
dpordomingo: we should maybe test the 
```
if record.GetID().IsEmpty() {
    return false, ErrEmptyID
}
```

store_test.go
```diff
@@ -169,6 +169,27 @@ func (s *StoreSuite) TestCount() {
...
+	s.Nil(rs.Scan(&model))
+
+	s.False(model.IsWritable())
+	s.Equal(0, model.Age)
+
+	s.Nil(s.store.Reload(&model))
+
+	s.True(model.IsWritable())
+	s.Equal(1, model.Age)
+}
```
Serabe: I don't understand the test and why sometimes this is writable and sometimes it is not.

store.go
```diff
@@ -211,6 +211,31 @@ func (s *Store) MustFind(q Query) *ResultSet {
 	return rs
 }
 
+// Reload refreshes the record with the data in the database and makes the
+// record writable.
+func (s *Store) Reload(record Record) error {
+	if record.GetID().IsEmpty() {
```
mcuadros: I know that is not related with this PR but ... `GetID`?  should be `ID`

dpordomingo: ID is already the name of a Model property
https://github.com/src-d/go-kallax/blob/base/model.go

dpordomingo: are you proposing (a) to rename the method, or (b) to access the property by name?
a) can not be done because ID is alredy used for the model property named `ID`
b) can not be accessed using .ID because Record is not an struct but an interface

mcuadros:
If your interface has a Getter out setter, in go a setter never has the
prefix Get

erizocosmico:
@mcuadros we know, but the `Model` struct already has a public field `ID`, that's why it's prefixed with `Get`. (It was also like that before in storable)

PR #33, opened 2017-01-20
Implement after hooks

erizocosmico:


generator/templates/base.tgo
```diff
@@ -9,5 +9,12 @@ import (
 
 var _ types.SQLType
 
+{{define "afterSaveEvent"}}err = s.Store.Transaction(func(store *kallax.Store) error {
```
erizocosmico: this is at the root because you can't use `define` inside child templates

PR #48, opened 2017-01-26
Automanage foreign keys and relationships

erizocosmico:


model.go
```diff
@@ -8,9 +8,10 @@ import (
 
 // Model is the base type of the items that are stored
 type Model struct {
-	ID        ID
-	persisted bool
-	writable  bool
+	ID             ID
+	virtualColumns []VirtualColumn
```
dpordomingo: Does the VirtualColumn need to be public?

erizocosmico: maybe not, i have to fix some stuff in there, so I will try to see if it can be private

erizocosmico:
Don't merge yet, needs a fix for some stuff

model.go
erizocosmico: `VirtualColumn` no longer exists. So now it's private :stuck_out_tongue:

erizocosmico:
Can be merged now

generator/templates/model.tgo
```diff
@@ -41,7 +41,10 @@ func (r *{{.Name}}) SetRelationship(field string, rel interface{}) error {
                 if !ok {
                         return fmt.Errorf("kallax: record of type %t can't be assigned to relationship {{.Name}}", rel)
                 }
-                r.{{.Name}} = {{if or (not .IsPtr) .IsOneToManyRelationship}}*{{end}}val
+                {{if .IsPtr}}if !val.Model.ID.IsEmpty() {
+                {{end}}r.{{.Name}} = {{if or (not .IsPtr) .IsOneToManyRelationship}}*{{end}}val{{if .IsPtr}}
+                }
+                {{end}}
```
Serabe: I think I rather see this two `if .IsPtr` as two branches of the same if. Something like

```go
{{if .IsPtr}}
if !val.Model.ID.IsEmpty() {
    r.{{.Name}} = {{if .IsOneToManyRelationship}}*{{end}}val
}
{{else}}
r.{{.Name}} = *val
{{end}}
```
Though slightly more verbose, much simpler to read. And I am starting to think Golang people did not think too much about its template syntax 😢 Wouldn't it be much simpler if you could indent your template in a way much simpler to read and then automagically use GoFmt on it?

generator/templates/model.tgo
```diff
@@ -87,7 +127,25 @@ func (s *{{.StoreName}}) Insert(record *{{.Name}}) error {
...
+                                return err
+                        }
+
+                        for _, r := range records {
+                                if _, err := s.Save(r.Schema, r.Record); err != nil {
+                                        return err
+                                }
+                        }
+
+                        return nil
```
Serabe: Review indentation, there are some lines using spaces instead of tabs

generator/templates/model.tgo
erizocosmico: `text/template` does weird things with blank on the `{{stuff}}`, so not much I can do on that side :(

But splitting this in two branches is nice, i'll change this asap

generator/templates/model.tgo
erizocosmico: I'll take a look at it

erizocosmico:
@Serabe done

tests/events.go
```diff
@@ -1,83 +1 @@
 package tests
-
```
dpordomingo: It was accidentally deleted `tests/events.go`

erizocosmico: nope, was not accidental. Was not compiling, so I wiped it haha

model.go
dpordomingo: super-private :+1:

tests/events.go
dpordomingo: it'll compile after https://github.com/src-d/go-kallax/pull/50

PR #58, opened 2017-01-31
Delete autogenerated kallax.go if generation failed

dpordomingo:


generator/generator.go
```diff
@@ -17,12 +20,20 @@ func (g *Generator) Generate(pkg *Package) error {
...
 	if err != nil {
 		return err
 	}
 
-	defer file.Close()
+	defer func() {
+		file.Close()
+		if err != nil {
+			fmt.Println()
+			fmt.Println("kallax: No file generated due to an occurred error:")
```
erizocosmico: remove first println and add a `\n` to the start of the second. Why the `:` if nothing else is being printed afterwards?

generator/generator.go
```diff
@@ -17,12 +20,20 @@ func (g *Generator) Generate(pkg *Package) error {
...
+			os.Remove(g.filename)
```
erizocosmico: this can fail

generator/generator.go
dpordomingo: this is an example of the error:
```
david@sourced: /projects/src/github.com/src-d/go-kallax$ kallaxgen 
Package:  tests
Found: "EventsFixture" [4 Field(s)] [Events: [BeforeInsert AfterInsert BeforeUpdate AfterUpdate]]
Found: "EventsSaveFixture" [4 Field(s)] [Events: [BeforeSave AfterSave]]
Found: "MultiKeySortFixture" [4 Field(s)] [Events: []]
Found: "QueryFixture" [2 Field(s)] [Events: []]
Found: "ResultSetFixture" [2 Field(s)] [Events: []]
Found: "SchemaFixture" [8 Field(s)] [Events: []]
Found: "StoreFixture" [2 Field(s)] [Events: []]
Found: "StoreWithConstructFixture" [2 Field(s)] [Events: []]
Found: "StoreWithNewFixture" [3 Field(s)] [Events: []]

kallax: No file generated due to an occurred error:
template: model:29:17: executing "model" at <0>: can't give argument to non-function 0
tests/common.go:3: running "kallax": exit status 1
```
After the `:` goes the error described by the line `due to an occurred error:`
Anyway, I can remove the `:` if you want.

Regarding to the `.Println()`, it is used to separate the generator log, and this new error (as a generation final error).
If we do:
`fmt.Println("\nkallax: No file generated due to an occurred error:")`
We break the rule: "_all errors must start by `kallax` keyword_"

dpordomingo:
@erizocosmico Done the second part of your suggestion; comment added to the first one

PR #73, opened 2017-02-06
Fix json schema init generation

erizocosmico:


ajnavarro:
This file has a lot of tests, but this line  (specifically, the complete "if") not. Could be a good idea to add a tests for this case?

erizocosmico:
Yes, we need to add tests for that.

El lun., 6 feb. 2017 a las 11:42, Antonio Navarro Perez (<notifications@github.com<mailto:notifications@github.com>>) escribió:

This file has a lot of tests, but this line (specifically, the complete "if") not. Could be a good idea to add a tests for this case?

PR #66, opened 2017-01-31
 Capture panics during generation, Write array length (issue#61), and other fixes

dpordomingo:
The only one feature added with this PR is:
- 6ed3026 Capture panics to do gracefully end of generation		

This PR fixes one known issue:
- 66039ee [Capture Array length to fix issue#61](https://github.com/src-d/go-kallax/issues/61)

Minor unrelated things:
- 3339697 Add TODO for already known issues
- ecae3cd FIX Outdated JSON integration tests

PR #75, opened 2017-02-06
FIX failing test

dpordomingo:
Update existent generator/types_test, thas was broken after https://github.com/src-d/go-kallax/pull/66


erizocosmico:
@dpordomingo you can undo your latest commit. It's not a race condition, it's a map relying on it being ordered (maps are not). I'll have it on my todo list for tomorrow, don't worry.

Anyway, if we ever have a race condition, I don't think removing parallel execution it's the right way to solve that :stuck_out_tongue:

PR #84, opened 2017-02-07
Make all struct fields embedded or defined inline inline

erizocosmico:


generator/processor.go
```diff
@@ -222,9 +222,17 @@ func (p *Processor) processFields(s *types.Struct, done []*types.Struct, root bo
...
 		}
 
+		var tname string
+		switch typ := f.Type().Underlying().(type) {
+		case *types.Struct:
+			tname = typeName(f.Type())
+		default:
+			tname = typeName(typ)
+		}
+
```
Serabe: I would extract this snippet to a `func` of its own.

erizocosmico:
@dpordomingo yes, this case has to be removed

generator/processor.go
erizocosmico: this has to be removed, don't worry

erizocosmico:
Now it can be reviewed again @Serabe @dpordomingo

PR #85, opened 2017-02-08
Document source code better and add pkg level comments

erizocosmico:


model.go
```diff
@@ -7,15 +7,35 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
-// Model is the base type of the items that are stored
+// Model contains all the basic fields that make something a model, that is,
+// the ID and some internal things used by kallax.
```
Serabe: s/internal things/internal data/

model.go
```diff
@@ -35,8 +56,7 @@ func (m *Model) setPersisted() {
 	m.persisted = true
 }
 
-// IsWritable returns whether this Model can be sent back to the database to be
-// stored with its changes.
+// IsWritable returns whether this Model can be saved into the database.
```
Serabe: I would add an example of why a model might not be writable.

model.go
```diff
@@ -96,17 +122,15 @@ type Persistable interface {
...
-	// IsWritable returns whether this Model can be sent back to the database
-	// to be stored with its changes.
+	// IsWritable returns whether this Model can be saved into the database.
 	IsWritable() bool
 	setWritable(bool)
 }
 
-// ColumnAddresser must be implemented by those values that expose their properties
-// under pointers, identified by its property names
+// ColumnAddresser provides the pointer addresses of columns.
```
dpordomingo: If this interface is not needed outside, it should be maybe made private.
If must be public for internal use, it should be explained in the doc;

Please, take this in consideration for the rest of the project :)

query.go
```diff
@@ -6,7 +6,9 @@ import (
 	"github.com/Masterminds/squirrel"
 )
 
-// Query returns information about some query settings and compiles the query.
+// Query is the common interface all queries must satisfy. The basic abilities
+// of a query are compiling themselves to something executable and return
+// some query settings.
```
dpordomingo: maybe private?

generator/types.go
```diff
@@ -429,6 +429,12 @@ type FieldKind int
 
 const (
 	// Basic is a field with a basic type.
+	// On top of the Go basic types, we consider Basic as well the following
+	// types:
+	//  - time.Time
+	//  - time.Duration
+	//  - kallax.ID
```
dpordomingo: are you sure about `kallax.ID`?

I think it matches this processor rule:
```
func (p *Processor) processField(field *Field, typ types.Type, ...) {
    switch typ := typ.(type) {
    ...
    case *types.Named:
        ...
        if p.isSQLType(types.NewPointer(typ)) {
            field.Kind = Interface
            return
        }
        ...
    ...
    }
}
```
so it would be an `Interface FieldKind`, isn't it?

erizocosmico:
Done @dpordomingo

PR #86, opened 2017-02-08
Add documentation in the README

erizocosmico:
Just the last commit, it's a chained PR.

If there is something we should change of the behaviour of kallax before v1, now would be the time to do so, so please, read carefully all the documentation and point out things you'd change before this makes it to the v1.

To read it formatted: https://github.com/erizocosmico/go-kallax/blob/docs/readme/README.md


README.md
```diff
@@ -15,6 +44,410 @@ go get -u github.com/src-d/kallax/...
...
+
+You can query arbitrary JSON using the JSON operators defined in the [kallax](https://godoc.org/github.com/src-d/go-kallax) package. The schema of the JSON (if it's a struct, obviously for maps it is not) is also generated.
+
+```go
+q := NewPostQuery().Where(kallax.JSONContainsAnyKey(
+        Schema.Post.Metadata,
+        "foo", "bar",
+))
+```
+
```
ajnavarro: Transactions?

README.md
```diff
@@ -15,6 +44,410 @@ go get -u github.com/src-d/kallax/...
 > *kallax* includes a binary tool used by [go generate](http://blog.golang.org/generate),
```
erizocosmico: `go get pkg/...` already installs

README.md
erizocosmico: Totally true

erizocosmico:
So, if we wanna release this to the public and not force them to use our UUIDs (which are actually ULIDs). Should we make the modification to allow primary keys of arbitrary types (simple ones, not composite keys) before v1.0. Because that would be a breaking change and we'd have to bump to v2.0 when the change is implemented.

@smola @ajnavarro @Serabe @dpordomingo @mcuadros

erizocosmico:
Done @ajnavarro

smola:
@erizocosmico As we discussed offline, we have not defined any IDs other than UUID/ULID so far, but defining our own for some future cases does not sound far fetched.

erizocosmico:
@smola yes, there is a PR for that (https://github.com/src-d/go-kallax/pull/98) and we'll update the docs once it's merged and the api for it is final. Right now, it is kind of a WIP to see if we like the solution implemented or not, so we're waiting for that before updating these docs

erizocosmico:
@dpordomingo I'm waiting for the custom pks thing to be merged before updating the docs in case something changes.

ajnavarro:
This documentation is actually outdated. Take into account the new way to implement the ID: `ID           kallax.ULID pk:""`

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+
+## Kallax generated code
+
+Kallax generates a bunch of code for every single model you have and saves it to a file named `kallax.go` in the same package.
+
+For every model you have, kallax will generate the following for you:
+
+* Internal methods for your model to make it work with kallax and satisfy the [Record](https://godoc.org/github.com/src-d/go-kallax#Record) interface.
+* A store named `{TypeName}Store`: the store is the way to access the data. A store of a given type is the way to access and manipulate data of that type.
+* A query named `{TypeName}Query`: the query is the way you will be able to build programmatically the queries to perform on the store. A store only will accept queries of its own type.
```
dpordomingo: I'd explain (here, or in the following query section), that adding filters to the query modifies the query itself (it was not obvious when we landed at squirrel for the first time and we needed to confirm it, so this clarification in advance would be useful for our new users)

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+
+So, to access the username field of the user model, it can be accessed as:
+
+```go
+Schema.User.Username
+```
+
+## Manipulate models
+
+For all of the following sections, we will assume we have a store `store` for our model's type.
```
dpordomingo: example?

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+
+### Insert models
+
+To insert a model we just need to use the `Insert` method of the store and pass it a model. If the model does not have an ID, one will be assigned to it.
```
dpordomingo: If the model does not have a primary key (not `ID`), and the defined primary key is not auto-incrementable, the insertion will fail.

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+if err != nil {
+        // handle error
+}
+```
+
+If there are any relationships in the model, both the model and the relationships will be saved in a transaction and only succeed if all of them are saved correctly.
+
+### Save models
+
+To save a model we just need to use the `Save` method of the store and pass it a model. It will update the model if it was already persisted or insert it otherwise.
```
dpordomingo: I'd explain that this method is just a shortcut to the proper `update` or `insert` depending on the previous state of the object, and I would explain no more. This way, we will avoid copy-paste in documentation, that can cause incoherence between the `save` and the main methods (example: it is not explained here that the save will fail if the object has not set its pk)

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+```
+
+## Query models
+
+### Simple queries
+
+To perform a query you have to do the following things: 
+* Create a query
+* Pass the query to `Find`, `FindOne`, `MustFind` or `MustFindOne` of the store
+* Gather the results from the result set
```
dpordomingo: FindOne and MustFindOne returns no result set but an object, or an error, or a panic

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+This process is repeated until there are no more rows in the result.
+Because of this, retrieving 1:N relationships is really fast.
+
+The default batch size is 50, you can change this using the `BatchSize` method all queries have.
+
+**NOTE:** if a filter is passed to a `With{Name}` method we can no longer guarantee that all related objects are there and, therefore, the retrieved records will **not** be writable.
+
+### Reloading a model
+
+If, for example, you have a model that is not writable because you only selected one field you can always reload it and have the full object.
```
dpordomingo: If you performed any change in an object before calling `reload`, the changes will be overridden with the db retrieved values

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+You can query just a single row with `FindOne`.
+
+```go
+q := NewUserQuery().
+        Where(kallax.Eq(Schema.User.Username, "Joe"))
+
+user, err := store.FindOne(q)
+```
+
+By default, all columns in a row are retrieved. To not retrieve all of them, you can specify the columns to include/exclude.
```
dpordomingo: I will explain that retrieving partial objects will make them not writable, and then I'd add a link to the `reload` section

dpordomingo:
- I'd add a caveats section, warning about the time aliases, and identifier aliases
- Could it be added the benchmark you made comparing kallax with gorm?
- I'd ad a logo or whatever image, random examples:
kallax
kallax2

whatever on your own ;)

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+        MetadataType MetadataType
+        Metadata map[string]interface{} // this will be json
+}
+```
+
+### Struct tags
+
+| Tag | Description | Can be used in |
+| --- | --- | --- | --- |
+| `table"table_name"` | Specifies the name of the table for a model | embedded `kallax.Model` |
```
erizocosmico: That's explained on other parts if I remember correclty

README.md
```diff
@@ -15,6 +47,507 @@ go get -u github.com/src-d/kallax/...
...
+| `pk:""` | Specifies the field is a primary key | any field with a valid identifier type |
+| `pk:"autoincr"` | Specifies the field is an auto-incrementable primary key | any field with a valid identifier type |
+| `kallax:"column_name"` | Specifies the name of the column | Any model field that is not a relationship |
+| `kallax:"-"` | Ignores the field and does not store it | Any model field |
+| `kallax:",inline"` | Adds the fields of the struct field to the model. Column name can also be given before the comma | Any struct field |
```
erizocosmico: The column name is ignored, because it is no longer a column

README.md
erizocosmico: what do you mean by example?

README.md
erizocosmico: That has an example afterwards, I think

erizocosmico:
The benchmark would need a little bit of work before being added here. And so would the logo haha. But yeah, we definitely need one.

README.md
erizocosmico: Actually, that part is no longer true and it needs to be removed.

README.md
dpordomingo: As far as I can see, it is explained the method used to generate the name of table columns and primary keys, but not for table names;
Anyway, I think that it could be great to define here (in this table, as a summary), the default values for each behavior, if the struct tag is not defined.

README.md
dpordomingo: now it is said: `Column name can also be given before the comma`, what is partially true;
it should be confirmed here in the readme, that if the inline contains more than one property, the column name before the `inline` tag will be ignored; otherwise the user could think that the provided column_name will be used for the column name generation in any way (as a prefix... or any other possible way)

README.md
dpordomingo: In the following examples we provide implementations for save, update... and we just use the store, but it is not said how to obtain it.
A possible example for this section could be:
```userStore := NewUserStore()```
Or just add the this store definition in the following save/update/insert examples

**Comment by erizocosmico** [comment]
@dpordomingo changes made

File: README.md (hunk 7 above)
Comment (dpordomingo) [inline README.md]: yes, there is a good example for the `FindOne`;

I'd add to the last step:
- _Gather the results from the result set **`returned by the Find and MustFind`**_

Do you think that it is unnecessary?

File: README.md [hunk 12]
```diff
@@ -186,7 +186,7 @@ Kallax will generate one with the following signature:
 func NewUser(username string, password string, emails ...string) (*User, error)
 ```
 
-**IMPORTANT:** if your primary key is not auto-incrementable, you should set an ID for every model you create in your constructor. Or, at least, set it before inserting it.
+**IMPORTANT:** if your primary key is not auto-incrementable, you should set an ID for every model you create in your constructor. Or, at least, set it before saving it. Inserting, updating, deleting or reloading an object with no primary key set will return an error.
```
erizocosmico: Yes, not always, but that block is only referring to "not auto-incrementable" pks.  `if your primary key is not auto-incrementable`

erizocosmico:
@dpordomingo where are the `//TO CONSIDER` comments? I can't find them

README.md
erizocosmico: Ok, we can define it there

README.md
erizocosmico: Ok, makes sense

README.md
erizocosmico: But it's explained on the `{TypeName}Store` section. Would it be necessary to add it there?

README.md
erizocosmico: Ok, it makes sense

erizocosmico:
Fixed @dpordomingo

README.md
dpordomingo: fair enough :+1:

README.md
dpordomingo: you're right :+1:

erizocosmico:
@dpordomingo added docs for findbys

README.md
```diff
@@ -15,6 +48,568 @@ go get -u github.com/src-d/kallax/...
...
+  * Types that implement `sql.Scanner` and `driver.Valuer`
+  The reason why this is not possible is because kallax implements support for arrays of all basic Go types by hand and also for types implementing `sql.Scanner` and `driver.Valuer` (using reflection in this case), but without having a common interface to operate on them, arbitrary types can not be supported.
+  For example, consider the following type `type Foo string`, using `[]Foo` would not be supported. Know that this will fail during the scanning of rows and not in code-generation time for now. In the future, might be moved to a warning or an error during code generation.
+  Aliases of slice types are supported, though. If we have `type Strings []string`, using `Strings` would be supported, as a cast like this `([]string)(&slice)` it's supported and `[]string` is supported.
+* `time.Time` and `url.URL` need to be used as is. That is, you can not use a type `Foo` being `type Foo time.Time`. `time.Time` and `url.URL` are types that are treated in a special way, if you do that, it would be the same as saying `type Foo struct { ... }` and kallax would no longer be able to identify the correct type.
+
```
dpordomingo: Arrays of arrays, and slices of lices (and any combination of these) are not supported by the db driver, so it can not be handled by Kallax.

erizocosmico:
Fixed @dpordomingo

Issue #39, opened 2017-01-25
Generate FindBys automatically for queries


dpordomingo:
tests in https://github.com/src-d/go-kallax/pull/50

erizocosmico:
Duplicated of https://github.com/src-d/go-kallax/issues/18

PR #92, opened 2017-02-13
Autogenerated FindBy

dpordomingo:
solves https://github.com/src-d/go-kallax/issues/39

## Functionality
Kallax will auto-generate the `FindBy` of each Model property if it satisfies one of the following;

A property will cause an auto-generated  `FindBy` if:
- its type a `"Valid"` one:
    - a) `"Basic"`, one of the following: `string`, `byte`, `bool`, `float`(s), `int`(s) or `uint`(s) ([see](https://github.com/src-d/go-kallax/pull/92/files#diff-4b6022e5bde395c01b628069f90029d8R142))
    - b) `"Special"` ([see](https://github.com/src-d/go-kallax/pull/92/files#diff-4b6022e5bde395c01b628069f90029d8R151))
        - one of the `"kallaxKnownTypes"` (SQL already knows how to handle it) ([see](https://github.com/src-d/go-kallax/pull/92/files#diff-404f36d691a324fb842e85a813e5735aR109))
        - a `"SQLType"` (implements `sql.Scanner` and `driver.Valuer`, [see](https://github.com/src-d/go-kallax/blob/master/generator/processor.go#L345))
- its type is a `"collection"` of `"Valid"` types

Type aliases will be valid if one of its underlying types is a `"Valid"` one.

## Generated FindBy
The auto-generated `FindBy` for a given property named `"PropertyName"` will be as it follows:

- `model.ID` -> `FindByID(v ...kallax.ID)`
the query will success if the model.ID is one of the passed `kallax.ID`
-  `"Basic"` and `"Special"` properties
    - `string` and `boolean` properties -> `FindByPopertyName(v validType)`,
  the query will success if the model property is `equal` to the passed `kallax.ID`
    - comparable properties -> `FindByPopertyName(cond kallax.ScalarCond, v validType)`,
  the query will success if the model property satisfies the passed `kallax.ScalarCond`
- `"Collection"` properties -> `FindByPopertyName(v ...singleValidType)`
the query will success if the model collection property contains one of the passed values


dpordomingo:
affected by https://github.com/src-d/go-kallax/issues/93

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"go/types"
+	"strings"
+)
+
+const (
+	tplFindByInclude = "\n\n// FindBy adds a new filter to the query that will require that\n" +
```
erizocosmico: we should use the ` for multiline strings instead of concatenated strings, it would be easier to read imho

generator/types.go
```diff
@@ -106,7 +106,7 @@ var reservedKeywords = map[string]struct{}{
 
 // special types that are not analyzed because SQL already knows
 // how to handle them
-var specialTypes = map[string]string{
+var kallaxKnownTypes = map[string]string{
```
erizocosmico: why the rename of this?

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	s = strings.Replace(s, "..", "", -1)
+	if requireShortName {
+		idx := strings.LastIndex(s, "/")
+		return s[1+idx:]
+	}
+
+	return s
+}
+
+func isEqualizable(f *Field) bool {
```
erizocosmico: some docs for these 3 functions would be nice for context

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return nil
+}
+
+// isBasicType returns true if passed type is one of the followings:
+// string, byte, bool, float(s), int(s) or uint(s)
+func isBasicType(pkg *types.Package, typ types.Type) bool {
+	s := typ.String()
+	return s == "string" || s == "bool" || s == "byte" || s == "float64" || s == "float32" ||
```
erizocosmico: what do you think about converting this to a switch, which would be cleaner?

```go
switch s {
case "string", "bool", "byte", "float64", ...:
    return true
}
return false
```

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+
+func getShortName(typ types.Type) string {
+	if special, ok := kallaxKnownTypeName(typ); ok {
+		return special
+	}
+
+	return cleanName(typ, true)
+}
+
+func cleanName(typ types.Type, requireShortName bool) string {
```
erizocosmico: this is already implemented https://github.com/src-d/go-kallax/blob/master/generator/types.go#L697

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	_, ok := kallaxKnownTypeName(typ)
+	return ok || isSQLType(pkg, typ)
+}
+
+func kallaxKnownTypeName(typ types.Type) (string, bool) {
+	special, ok := kallaxKnownTypes[cleanName(typ, false)]
+	return special, ok
+}
+
+func getShortName(typ types.Type) string {
```
erizocosmico: what do you think about naming this `shortName`, i feel `get` is very redundant here

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	query := parent.QueryName
+	model := parent.Name
+	buf.WriteString(fmt.Sprintf(tpl, property, query, validType, model))
+}
+
+// validType returns the short name of the type that can be used to search for
+// the passed Field, in an autogenerated 'FindBy';
+// if the Field is not valid for an autogenerated 'FindBy' it will return an error
+func validType(f *Field) (string, error) {
+	analycedType := f.Node.Type()
```
erizocosmico: `analyzedType`, but imho, `t` it's just fine for this

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+		analycedType = singular
+		collectionAlreadyScanned = true
+	}
+
+	return "", fmt.Errorf("kallax: '%s' is not valid for FindBy", f.Node.Type())
+}
+
+// lookupValid returns the first valid type looking into the underlying types of
+// the passed one; if no valid found it will return the deepest underlying
+func lookupValidRecursively(pkg *types.Package, typ types.Type) (valid, deepest types.Type) {
```
erizocosmico: i'd remove the `Recursively` from the name. That is implementation details and has nothing to do with the name

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		analycedType = singular
+		collectionAlreadyScanned = true
+	}
+
+	return "", fmt.Errorf("kallax: '%s' is not valid for FindBy", f.Node.Type())
+}
+
+// lookupValid returns the first valid type looking into the underlying types of
```
erizocosmico: wrong func name

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+// the passed one; if no valid found it will return the deepest underlying
+func lookupValidRecursively(pkg *types.Package, typ types.Type) (valid, deepest types.Type) {
+	for _, isValid := range typeValidators {
+		if isValid(pkg, typ) {
+			return typ, nil
+		}
+	}
+
+	underlaying := typ.Underlying()
```
erizocosmico: `underlaying` => `underlying`

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+		collectionAlreadyScanned = true
+	}
+
+	return "", fmt.Errorf("kallax: '%s' is not valid for FindBy", f.Node.Type())
+}
+
+// lookupValid returns the first valid type looking into the underlying types of
+// the passed one; if no valid found it will return the deepest underlying
+func lookupValidRecursively(pkg *types.Package, typ types.Type) (valid, deepest types.Type) {
+	for _, isValid := range typeValidators {
```
erizocosmico: it's just two cases, i'd remove the `typeValidators` slice and do a 

```go
if isBasicType(t) || isSpecialType(t)
```

which is more explicit and for sure we won't have more of them

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	case typ != underlaying:
+		return lookupValidRecursively(pkg, underlaying)
+	default:
+		return nil, typ
+	}
+}
+
+// singularType returns the type of the elements of the passed collection,
+// or nill if the passed type is not a collection
+func singularType(typ types.Type) types.Type {
```
erizocosmico: what do you think about naming this `collectionType` or `collectionElemType`?

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	switch typ.(type) {
+	case *types.Array:
+		return typ.(*types.Array).Elem()
```
erizocosmico: if you make `switch typ := typ.(type)` you can save the type cast

generator/findby.go
```diff
@@ -0,0 +1,192 @@
+package generator
```
erizocosmico: this should be in the `template.go` file

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+		"func (q *%[2]s) FindBy%[1]s(v ...%[3]s) *%[2]s {\n" +
+		"    if len(v) == 0 {return q}\n" +
+		"    values := make([]interface{}, len(v))\n" +
+		"    for i, val := range v {values[i] = val}\n" +
+		"    return q.Where(kallax.In(Schema.%[4]s.%[1]s, values...))\n" +
+		"}"
+)
+
+// isValid returns true if the passed type is valid for an autogenerated FindBy
+type isValid func(pkg *types.Package, typ types.Type) bool
```
erizocosmico: if we remove `typeValidators`, this would be pointless to keep

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+			writeTplCode(buf, parent, f, tplFindByCondition)
+		case isCollection(f):
+			writeTplCode(buf, parent, f, tplFindByInclude)
+		}
+	}
+}
+
+func writeTplCode(buf *bytes.Buffer, parent *Model, f *Field, tpl string) {
+	validType, err := validType(f)
+	if err != nil {
```
erizocosmico: if the error it's not going to be treated as an error, return a boolean instead

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+	property := f.Name
+	query := parent.QueryName
+	model := parent.Name
+	buf.WriteString(fmt.Sprintf(tpl, property, query, validType, model))
+}
+
+// validType returns the short name of the type that can be used to search for
+// the passed Field, in an autogenerated 'FindBy';
+// if the Field is not valid for an autogenerated 'FindBy' it will return an error
+func validType(f *Field) (string, error) {
```
erizocosmico: what do you think about naming this `findableTypeName`?

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// isValid returns true if the passed type is valid for an autogenerated FindBy
+type isValid func(pkg *types.Package, typ types.Type) bool
+
+var typeValidators = []isValid{isBasicType, isSpecialType}
+
+// GenFindBy generates FindByPropertyName for all model properties that are
+// valid types or collection of valid types.
+func (td *TemplateData) GenFindBy(model *Model) string {
```
erizocosmico: this should have tests

generator/findby.go
erizocosmico: this should have tests

generator/findby.go
erizocosmico: this should have tests

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// isSpecialType returns true if the passed type is one of the kallaxKnownTypes
+// or a types.SQLType
+func isSpecialType(pkg *types.Package, typ types.Type) bool {
+	_, ok := kallaxKnownTypeName(typ)
+	return ok || isSQLType(pkg, typ)
+}
+
+func kallaxKnownTypeName(typ types.Type) (string, bool) {
```
erizocosmico: if the name of `kallaxKnownTypes` changes, this should change too

generator/findby.go
```diff
@@ -0,0 +1,192 @@
...
+func isBasicType(pkg *types.Package, typ types.Type) bool {
+	s := typ.String()
+	return s == "string" || s == "bool" || s == "byte" || s == "float64" || s == "float32" ||
+		s == "int" || s == "int8" || s == "int16" || s == "int32" || s == "int64" ||
+		s == "uint" || s == "uint8" || s == "uint16" || s == "uint32" || s == "uint64"
+}
+
+// isSpecialType returns true if the passed type is one of the kallaxKnownTypes
+// or a types.SQLType
+func isSpecialType(pkg *types.Package, typ types.Type) bool {
```
erizocosmico: if `kallaxKnownTypes` changes, this should change too

generator/findby.go
erizocosmico: this should have tests

generator/findby.go
erizocosmico: this needs tests

tests/query_test.go
```diff
@@ -40,3 +88,115 @@ func (s *QuerySuite) TestQuery() {
...
+	s.NotPanics(func() {
+		count, err := store.Count(NewQueryFixtureQuery().FindByInteger(kallax.Lt, 0))
+		s.Equal(int64(0), count)
+		s.Nil(err)
+	})
+}
+
+func (s *QuerySuite) TestGeneration() {
+	q := NewQueryFixtureQuery()
+	s.hasFindByMethod(q, "ID")
```
erizocosmico: how about a trable-driven tests?

```go
var cases = []struct{
    name string
    generated bool
}{
    {"ID", true},
    // ...
}

for _, c := range cases {
    if c.generated {
        s.hasFindByMethod(q, c.name)
    } else {
        s.hasNotFindByMethod(q, c.name)
    }
}
```

that way it's easier to add more cases and there is less repeated code

generator/findby.go
dpordomingo: much better

generator/findby.go
dpordomingo: good idea ;)

generator/findby.go
dpordomingo: #epicFail, I just used the right word at the right
`underlaying := typ.Underlying()`

generator/findby.go
dpordomingo: I'd rather prefer the second one: `collectionElemType`
The first one seems to be _"the type of the collection"_

generator/findby.go
dpordomingo: tricky... but ok :)

generator/findby.go
dpordomingo: did't changed :)

generator/findby.go
dpordomingo: did't changed :)

generator/findby.go
dpordomingo: I forgot to remove this getter, thanks for point  out this one

generator/findby.go
```diff
@@ -158,14 +162,15 @@ func specialTypeShortName(typ types.Type) (string, bool) {
...
-func getShortName(typ types.Type) string {
+func shortName(typ types.Type) string {
 	if special, ok := specialTypeShortName(typ); ok {
 		return special
 	}
 
 	return cleanName(typ, true)
 }
 
+//TODO: review existent implementation generator.types::typeString
```
erizocosmico: space after `//`

erizocosmico:
Btw, a lot of tests are failing

dpordomingo:
@erizocosmico many thanks for your recommendations!!!
I applied most of your suggested changes but:
- **moving `findby.go` into `template.go`** in order to make it easier the sequential review of the PR,
- **adding tests**, that will be done from now on, since we agree in the api and global behavior,
- **using generator.typeString instead of generator.cleanName**, after the tests (of the previous step) has been done, to ensure that this changes breaks nothing.

dpordomingo:
affected by https://github.com/src-d/go-kallax/issues/96

erizocosmico:
Is this ready for review again? @dpordomingo

dpordomingo:
I forgot pushing it;
58877ad Use generator.types::typeString as suggested by @erizocosmico
a11cfb2 Add test for autogenerated FindBy			

Pending, and will be done after your validation, in order to make it easier the sequential review of the PR :
- **moving `findby.go` into `template.go`** in order to make it easier the sequential review of the PR,
- **squash and rebase** after your validation

erizocosmico:
@dpordomingo so is it ready for review, right?

dpordomingo:
Yes, it is :)

tests/query_test.go
```diff
@@ -40,3 +89,124 @@ func (s *QuerySuite) TestQuery() {
...
+	store := NewQueryFixtureStore(s.db)
+	s.NotPanics(func() {
+		//s.True(store.MustFindOne(NewQueryFixtureQuery().FindByStringProperty("StringProperty1")).Eq(queryFixtures[1]))
+	})
+	s.Panics(func() {
+		store.MustFindOne(NewQueryFixtureQuery().FindByStringProperty("NOT_FOUND"))
+	})
+
+	s.NotPanics(func() {
+		//s.True(store.MustFindOne(NewQueryFixtureQuery().FindByBoolean(false)).Eq(queryFixtures[1]))
```
erizocosmico: leftovers?

dpordomingo: nope,
Those tests are already broken tests, affected by https://github.com/src-d/go-kallax/issues/96

I updated it in my rebased branch, and it looks this way:
image

erizocosmico: Oh, ok, then lgtm

Serabe:
@dpordomingo tests are failing. Are they supposed to fail yet?

dpordomingo:
@erizocosmico @Serabe Tests now passes, it was needed to use the go-parse-utils importer 58a187c47ec35abf3af7311cf3556ef9cb5b7b3c

generator/template.go
```diff
@@ -429,3 +430,198 @@ var resultset *template.Template = addTemplate(model, "resultset", "templates/re
...
+	for _, f := range fields {
+		switch {
+		case f.Inline():
+			td.genFindBy(buf, parent, f.Fields)
+		case f.Name == "ID":
+			writeTplCode(buf, parent, f, tplFindByID)
+		case isEqualizable(f):
+			writeTplCode(buf, parent, f, tplFindByEquality)
+		case isSortable(f):
+			writeTplCode(buf, parent, f, tplFindByCondition)
```
Serabe: This path is not tested.

Issue #93, opened 2017-02-13
sql.Error when using []struct{} properties


dpordomingo:
```
Error:		Received unexpected error:
sql: Scan error on column index 3: kallax: cannot scan type %!s(<nil>) into JSON type
```
test in https://github.com/src-d/go-kallax/pull/94

PR #94, opened 2017-02-13
Test issue#93. Error in []Struct properties

dpordomingo:
described by https://github.com/src-d/go-kallax/issues/93

### DIFF
--- a/tests/kallax.go
+++ b/tests/kallax.go
@@ -1768,7 +1768,7 @@ func (r *JSONModel) Value(col string) (interface{}, error) {
 	case "bar":
 		return types.JSON(r.Bar), nil
 	case "baz_slice":
-		return types.Slice(r.BazSlice), nil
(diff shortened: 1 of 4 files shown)


tests/kallax.go
```diff
@@ -1765,6 +1767,8 @@ func (r *JSONModel) Value(col string) (interface{}, error) {
 		return r.Foo, nil
 	case "bar":
 		return types.JSON(r.Bar), nil
+	case "baz_slice":
+		return types.Slice(r.BazSlice), nil
```
erizocosmico: the problem might be this. It should be `types.JSON`, not `types.Slice`

dpordomingo: Yes, it is!!
now, lets see how to fix that autogenerated code :D

dpordomingo:
I think it's now solved @erizocosmico

PR #98, opened 2017-02-15
Implement customizable primary key types

erizocosmico:
Still a WIP, missing tests and all, but just to validate the API and the changes.

(fear not, most of the changes are generated kallax.go changes)

### Considerations

Is not totally customizable, only these types are allowed:
- `int64`
- `uuid.UUID`
- `kallax.UUID`
- `kallax.ULID`
- `kallax.NumericID`

### DIFF
--- a/generator/types.go
+++ b/generator/types.go
@@ -124,7 +124,30 @@ type Package struct {
 	// Name is the package name.
 	Name string
 	// Models are all the models found in the package.
-	Models []*Model
+	Models        []*Model
+	indexedModels map[string]*Model
+}
+
+// NewPackage creates a new package.
+func NewPackage(pkg *types.Package) *Package {
+	return &Package{
+		Name:          pkg.Name(),
+		pkg:           pkg,
+		indexedModels: make(map[string]*Model),
+	}
+}
+
+// SetModels sets the models of the packages and indexes them.
+func (p *Package) SetModels(models []*Model) {
+	for _, m := range models {
+		p.indexedModels[m.Name] = m
+	}
+	p.Models = models
+}
+
+// FindModel finds the model with the given name.
+func (p *Package) FindModel(name string) *Model {
+	return p.indexedModels[name]
 }
 
 const (
@@ -156,6 +179,8 @@ type Model struct {
 	Type string
 	// Fields contains the list of fields in the model.
 	Fields []*Field
+	// ID contains the identifier field of the model.
+	ID *Field
 	// Events contains the list of events implemented by the model.
 	Events Events
 	// Node is the node where the model was defined.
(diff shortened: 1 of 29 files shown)


generator/types.go
```diff
@@ -368,11 +401,30 @@ func (m *Model) CtorRetVars() string {
...
+		if f.Type == BaseModel {
+			m.ID = f.Fields[0]
+			m.ID.Type = defaultIDType
+			id = f
+		}
+
 		f.Model = m
-		m.Fields = append(m.Fields, f)
+		if f.IsPrimaryKey() {
+			id = f
```
dpordomingo: Is it guarantied in any other place that there is only one field that f.IsPrimaryKey() == true?
If we are not supported compounde PK, we should maybe "panic" if the user tries to define a compounded one

tests/kallax.go
```diff
@@ -3120,29 +3162,32 @@ func (rs *PersonResultSet) Close() error {
 
 // NewPet returns a new instance of Pet.
 func NewPet(name string, kind string, owner *Person) (record *Pet) {
-	record = newPet(name, kind, owner)
-	if record != nil {
-		record.SetID(kallax.NewID())
-	}
-	return
+	return newPet(name, kind, owner)
```
dpordomingo: I think that every brand new model that `!schema.autoIncr` should have its own ID because Model fulfill the `IDSetter` interface, shouldn't they?

tests/store_test.go
```diff
@@ -40,7 +40,7 @@ type StoreSuite struct {
 func (s *StoreSuite) TestStoreNew() {
 	doc := NewStoreWithConstructFixture("foo")
 	s.False(doc.IsPersisted())
-	s.False(doc.ID.IsEmpty())
+	s.True(doc.ID.IsEmpty())
```
dpordomingo: If every Model implements `IDSetter`... then this should be maintained, shouldn't it?

generator/types.go
erizocosmico: In this very method the pk is only added once. So if there is any other PK is not added as field.

tests/kallax.go
erizocosmico: that's for the user to implement, because we don't know what they want as value in there

tests/store_test.go
erizocosmico: Yes, but it is not added in the `New{foo}` method, but when you are going to insert just as a last resort.

If you need to have an ID, it's your job to provide one using the private constructor. That should be in the docs, cause it can be misleading.

generator/types.go
dpordomingo: let me explain it...
current implementation is:
```
func (m *Model) SetFields(fields []*Field) {
	var id *Field
	for _, f := range fields {
		...
		if f.IsPrimaryKey() {
			id = f
			m.ID = f
		} 
		...
	}
	...
}
```
If the receiver Model has two fields identified as `"kallax:,pk"`, the second one will override the first one.
I think that, in that case (two defined `pk` in the same model), the kallax generation should panic

tests/kallax.go
dpordomingo: it can be solved just calling the `record.SetID()` in the record constructor (done in the generated `kallax.go`)
Anyway, that's what we were talking about, and we agreed on:
- user implements `newRecord()`
- then `kallax.go` will ensure that NewRecord() calls the `newRecord()`
Am I right?

tests/store_test.go
dpordomingo: I think that in this PR, the old test assertion should be maintained, implementing the "ID generation" in the fixture `newStoreWithConstructFixture(f string) *StoreWithConstructFixture` method, to ensure that the tested behavior is maintained;

erizocosmico:
@dpordomingo @Serabe removed ID from Model

generator/types.go
erizocosmico: Actually, yes. It should.

tests/kallax.go
erizocosmico: yes, that's what happened

generator/types.go
erizocosmico: Done @dpordomingo

PR #100, opened 2017-02-16
cli: add `exclude` flag to exclude files from generation

erizocosmico:


generator/cli/kallax/cmd.go
```diff
@@ -20,12 +20,16 @@ func main() {
...
 		},
 		cli.StringFlag{
 			Name:  "output",
 			Value: "kallax.go",
-			Usage: "output file name",
+			Usage: "Output file name",
+		},
+		cli.StringSliceFlag{
+			Name:  "exclude, e",
+			Usage: "List of excluded files in the excluded. You can use this flag as many times as you want.",
```
Serabe: I don't understand _List of excluded files in the excluded_

PR #101, opened 2017-02-17
fix: ensure nullable types work correctly with pointer types

erizocosmico:
So, how to do that?

By implementing again all basic types in Nullable, only this time not for `*T` but also for `**T`, which allows you to set a value even if the ptr was initially `nil`.

This fixes the bug @ajnavarro reported yesterday, there's a test now to ensure that.

### DIFF
--- a/generator/template.go
+++ b/generator/template.go
@@ -73,7 +73,7 @@ func (td *TemplateData) genFieldsColumnAddresses(buf *bytes.Buffer, fields []*Fi
 				if f.IsJSON && f.IsPtr {
 					buf.WriteString(fmt.Sprintf(initNilPtrTpl, f.Name, f.Name, td.GenTypeName(f)))
 				}
-				buf.WriteString(fmt.Sprintf("return %s\n", f.Address()))
+				buf.WriteString(fmt.Sprintf("return %s, nil\n", f.Address()))
 			}
 		}
 	}
--- a/generator/template.go
+++ b/generator/template.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"go/build"
+	"go/types"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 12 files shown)


types/types.go
```diff
@@ -391,6 +800,8 @@ func (j *sqlJSON) Scan(v interface{}) error {
 		return json.Unmarshal(v, j.val)
 	case string:
 		return j.Scan([]byte(v))
+	case nil:
+		return nil
```
ajnavarro: You are not testing this new case.

generator/types.go
```diff
@@ -677,27 +677,36 @@ func (f *Field) fieldVarAddress() string {
...
 
 	if f.Kind == Array {
-		return fmt.Sprintf("types.Array(%s, %d), nil", ptr, arrayLen(f))
+		return fmt.Sprintf("types.Array(%s, %d)", ptr, arrayLen(f))
 	}
 
-	return fmt.Sprintf("%s, nil", ptr)
+	if f.IsPtr && !casted {
+		if f.Kind == Interface {
+			return fmt.Sprintf("types.Nullable(%s)", ptr)
```
ajnavarro: Test this too.

types/types.go
```diff
@@ -64,6 +64,38 @@ func Nullable(typ interface{}) sql.Scanner {
...
+		return &nullPtrFloat32{typ}
+	case **float64:
+		return &nullPtrFloat64{typ}
+	case **time.Time:
+		return &nullPtrTime{typ}
+	case **time.Duration:
+		return &nullPtrDuration{typ}
 	}
 
 	return &nullableErr{typ}
```
ajnavarro: You can test this with a type that is not into the switch case.

generator/template.go
```diff
@@ -70,10 +71,10 @@ func (td *TemplateData) genFieldsColumnAddresses(buf *bytes.Buffer, fields []*Fi
 				buf.WriteString(fmt.Sprintf("return (*%s)(%s), nil\n", td.IdentifierType(f), f.fieldVarAddress()))
 			} else {
 				// can't scan a json if is nil
-				if f.IsJSON && f.IsPtr {
+				if (f.IsJSON || f.Kind == Interface) && f.IsPtr {
 					buf.WriteString(fmt.Sprintf(initNilPtrTpl, f.Name, f.Name, td.GenTypeName(f)))
```
erizocosmico: yes, it does https://github.com/src-d/go-kallax/pull/101/files#diff-35b32f9431b138f0c3a4d7b2dc2c8fc7R117

types/types.go
erizocosmico: if a type is not on the switch, it's a `*nullableErr`, which is just a type that will return an error either using `Value` or `Scan`

generator/template.go
ajnavarro: I just saw this: https://codecov.io/gh/src-d/go-kallax/pulls/101/diff#D6-75

erizocosmico:
@ajnavarro done

PR #102, opened 2017-02-17
Bugfix/scan slice alias

### LINKED ISSUE #96: store.Find fails when the struct has a property that is an alias of a slice
store.Find fails when the struct has a property that is an alias of a slice.
```
Error:		func (assert.PanicTestFunc)(0x4a9080) should not panic
Panic value:	sql: Scan error on column index 3: pq: parsing array element index 0: pq: scanning to string is not implemented; only sql.Scanner
```
If the struct has contain (not `empty` content) under a property that is an Alias of a slice, any Find that reach that struct will fail.
If the property is empty, the Find will not fail.

Tests in PR https://github.com/src-d/go-kallax/pull/97

erizocosmico:
Fixes #96

### DIFF
--- a/batcher_test.go
+++ b/batcher_test.go
@@ -21,7 +21,7 @@ func TestOneToManyWithFilterNotWritable(t *testing.T) {
 	r.NoError(store.Insert(ModelSchema, m))
 
 	for i := 0; i < 4; i++ {
-		r.NoError(store.Insert(RelSchema, newRel(m.ID, fmt.Sprint(i))))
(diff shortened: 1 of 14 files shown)


erizocosmico:
Don't merge yet, I discovered it needs a bit more of work before being fully complete.

Serabe:
Please, ping when it is ready.

erizocosmico:
Will do

erizocosmico:
Can be reviewed now @dpordomingo @Serabe

PR #103, opened 2017-02-20
Find by leftovers

dpordomingo:
Removing leftovers, improving code coverage in FindBy autogenerated methods and finxing a problem with Array

### DIFF
--- a/generator/template.go
+++ b/generator/template.go
@@ -623,17 +623,19 @@ func specialTypeShortName(typ types.Type) (string, bool) {
 }
 
 func shortName(pkg *types.Package, typ types.Type) string {
-	var preffix string
+	var prefix string
 	if singleType := collectionElemType(typ); singleType != nil {
+		t := typ.String()
+		idx := strings.Index(t, "]")
+		prefix = t[:idx+1]
 		typ = singleType
-		preffix = "[]"
 	}
 
 	if specialName, ok := specialTypeShortName(typ); ok {
-		return preffix + specialName
+		return prefix + specialName
 	} else {
 		shortName := typeString(typ, pkg)
-		return preffix + strings.Replace(shortName, "*", "", -1)
+		return prefix + strings.Replace(shortName, "*", "", -1)
 	}
 }
 
--- a/generator/template_test.go
+++ b/generator/template_test.go
@@ -429,7 +429,8 @@ func (s *ProcessorSuite) TestFindableTypeName() {
 			ID						kallax.ULID			` + "`findable:\"kallax.ULID\" pk:\"\"`" + `
 
 			StringProp 				string 				` + "`findable:\"string\"`" + `
-			ArrStringProp 			[]string 			` + "`findable:\"string\"`" + `
+			SliceStringProp 		[]string 			` + "`findable:\"string\"`" + `
+			ArrStringProp 			[2]string 			` + "`findable:\"string\"`" + `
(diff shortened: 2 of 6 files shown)


operators.go
```diff
@@ -11,22 +11,24 @@ import (
...
+// ScalarCond returns a kallax.Condition that compares a property with the passed
+// values, considering its scalar values (eq, gt, gte, lt, lte, neq)
+type ScalarCond func(col SchemaField, n interface{}) Condition
 
 // Condition represents a condition of filtering in a query.
 type Condition func(Schema) squirrel.Sqlizer
 
 // Eq returns a condition that will be true when `col` is equal to `value`.
-func Eq(col SchemaField, value interface{}) Condition {
+func Eq(col SchemaField, n interface{}) Condition {
```
erizocosmico: I would not change this to `n` as this implies it's a number, but it's not. `Eq` can be used with almost all types.

operators.go
```diff
@@ -11,22 +11,24 @@ import (
...
 	return func(schema Schema) squirrel.Sqlizer {
-		return squirrel.Eq{col.QualifiedName(schema): value}
+		return squirrel.Eq{col.QualifiedName(schema): n}
 	}
 }
 
 // Lt returns a condition that will be true when `col` is lower than `value`.
-func Lt(col SchemaField, value interface{}) Condition {
+func Lt(col SchemaField, n interface{}) Condition {
```
erizocosmico: ditto

generator/template_test.go
```diff
@@ -91,6 +93,7 @@ const baseTpl = `
 		ID int64 ` + "`pk:\"autoincr\"`" + `
 		Foo string
 		Bar *string
+		Baz int64
```
erizocosmico: Why was an `int64` field added?

dpordomingo: it increase the coverage of the FindBy because now it is considered the sortable properties

operators.go
dpordomingo: It was done for consistency with the other `ScalarCond` functions: `Gt`, `LtOrEq`, `GtOrEq` and `Neq`, that uses `n` as the second function parameter.
If you preffer `value` name, I can change it in the other listed above.

dpordomingo:
@erizocosmico questions answered

generator/template.go
```diff
@@ -625,8 +625,10 @@ func specialTypeShortName(typ types.Type) (string, bool) {
 func shortName(pkg *types.Package, typ types.Type) string {
 	var preffix string
 	if singleType := collectionElemType(typ); singleType != nil {
+		t := typ.String()
+		idx := strings.Index(t, "]")
+		preffix = t[:idx+1]
```
erizocosmico: preffix should be prefix

generator/template_test.go
erizocosmico: oh, ok

operators.go
erizocosmico: then we should change them all to something that is consistent but also is not `n`, `v` or `value`, for example

dpordomingo: done  d64450a

erizocosmico:
Please, squash before merging and after this we can release v1.0

dpordomingo:
@erizocosmico I rebased the branch.
Your requested changes are in:
- d64450a Use 'value' param name as required by @erizocosmico
- e98f297 Fix a typo sptted by @erizocosmico

PR #119, opened 2017-04-19
Fixing "Struct tags" table in the README

MohamedBassem:
That's a minor README change.


erizocosmico:
Thanks for the contribution! Nice catch!

PR #126, opened 2017-04-20
save/update inverses before record

### LINKED ISSUE #117: Save/update order in inverse relationships
Right now, the entity is saved first and then all its relationships. This obviously does not work for inverses, because the entity depends on its inverse relationships.

It would just be a matter of changing the order of insertion/save to the following:

1. Inverses
2. Entity
3. Relationships

erizocosmico:
Fixes #117 

Before, inverses were being treated as any other relationship, thus being saved/updated after the model. If the model had any inverses, that had not been previously saved, the operation would fail, as the inverse is not saved but required to exist in the database.
Now, inverses are saved first, then the model, and, finally, the other relationships.

### DIFF
--- a/generator/templates/model.tgo
+++ b/generator/templates/model.tgo
@@ -96,26 +96,17 @@ func New{{.StoreName}}(db *sql.DB) *{{.StoreName}} {
 	return &{{.StoreName}}{kallax.NewStore(db)}
 }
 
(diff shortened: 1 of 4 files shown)


erizocosmico:
@smola the transaction is automatically done by kallax if it has to save more than one record (the case of relationships, for example)

PR #127, opened 2017-04-20
implement StoreFrom function to transmute stores

### LINKED ISSUE #118: Get typed stores of all relationships from a typed store
Right now, to convert from a typed store to another, you have to do the following:

```go
s1 := NewT1Store(db)
s2 := &T2Store{s1.Store}
```

That requires you to know that a typed store is a struct with an embedded untyped store, which is not something the user should be required to know.

It would be nice to be able to do something like:

```go
s1 := NewT1Store(db)
s2 := s1.T2Store()
```

Maybe, it would be nice to add all possible store conversions to all stores and not just related ones.

erizocosmico:
Fixes #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,11 +572,12 @@ store.Transaction(func(s *UserStore) error {
 })
 ```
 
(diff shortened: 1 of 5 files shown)


Serabe:
Build is broken though

PR #133, opened 2017-04-23
Fix import paths using windows path separators (fixes #129)

### LINKED ISSUE #129: Kallax gen isn't picking up any models
$ go version
go version go1.8 windows/amd64

$ kallax -v
kallax version 1.0.0

When I run `$ kallax gen` it creates a generic kallax.go file. I'm running it in the tests dir of this project.
```go
// IMPORTANT! This is auto generated code by https://github.com/src-d/go-kallax
// Please, do not touch the code below, and if you do, do it under your own
// risk. Take into account that all the code you write here will be completely
// erased from earth the next time you generate the kallax models.
package tests

import (
        "fmt"

        "gopkg.in/src-d/go-kallax.v1/types"
)

var _ types.SQLType
var _ fmt.Formatter

type schema struct {
}

var Schema = &schema{}
```

I did run go get -u gopkg.in/src-d/go-kallax.v1/... and I did not get v1.1.0.

JaCoB1123:
Convert back-slashes in import path to slashes so the comparison to the base type succeeds.

### DIFF
--- a/generator/processor.go
+++ b/generator/processor.go
@@ -466,7 +466,9 @@ func typeName(typ types.Type) string {
 }
 
 func removeGoPath(path string) string {
-	return strings.Replace(path, goPath+"/src/", "", -1)
+	importPath := filepath.ToSlash(goPath + "/src/")
+	path = filepath.ToSlash(path)
+	return strings.Replace(path, importPath, "", -1)
 }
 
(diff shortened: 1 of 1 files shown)


erizocosmico:
Thanks for the contribution!

PR #134, opened 2017-04-23
No way to override this within my model defs

gen0cide-:
I'd prefer to keep my JSON keys snake case, and afaik, there was no way to override this easily without loosing the Timestamps functionality.

### DIFF
--- a/timestamps.go
+++ b/timestamps.go
@@ -14,9 +14,9 @@ import "time"
 //	}
 type Timestamps struct {
 	// CreatedAt is the time where the object was created.
-	CreatedAt time.Time
+	CreatedAt time.Time `json:"created_at"`
 	// UpdatedAt is the time where the object was updated.
-	UpdatedAt time.Time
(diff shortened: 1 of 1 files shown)


erizocosmico:
Thanks for the contribution. Nevertheless, we can not merge this as it is. If this does not offer a general way to change the names of the fields, it's just a matter of "user A prefers UpperCamelCase" and "user B prefers snake_case". If we can only have one name for the fields, we prefer it to be the default.

You can always use your custom `Timestamps` on your app if you really need this.

```go
type Timestamps struct {
  CreatedAt  time.Time `json:"created_at"`
  UpdatedAt time.Time `json:"updated_at"`
}

func (t *Timestamps) BeforeSave() error {
  if t.CreatedAt.IsZero() {
    t.CreatedAt = time.Now()
  }
  t.UpdatedAt = time.Now()
  return nil
}
```

There's nothing special in the library's `Timestamps`, so it would work the exact same way.

PR #135, opened 2017-04-23
always use the proxy in all store operations instead of raw database

### LINKED ISSUE #132: Running `Find` inside Transaction will panic due to `s.db` being `nil`
Reproduce

``` go
// store_test.go

func (s *StoreSuite) TestTransaction_Query() {
	err := s.store.Transaction(func(store *Store) error {
		q := NewBaseQuery(ModelSchema)
		q.Select(NewSchemaField("name"), ModelSchema.ID())
		_, err := store.Find(q)
		s.Nil(err)
		return nil
	})
	s.Nil(err)
}
```

The problem (I think) is that `newStoreWithTransaction` returns a `*Store` with `proxy` and `builder` set, `db` empty. Then in places like `Find`, `RawQuery` and `RawUpdate`, uses `s.db` directly.

I don't know `squirrel` enough, but can we simply swap out `s.db` with `s.proxy` instead?

erizocosmico:
Fixes #132 

transactions have a nil database, so they only have the proxy. Using
the database directly makes the operations fail in transactions.

### DIFF
--- a/store.go
+++ b/store.go
@@ -217,7 +217,7 @@ func (s *Store) Delete(schema Schema, record Record) error {
 // WARNING: A result set created from a raw query can only be scanned using the
 // RawScan method of ResultSet, instead of Scan.
 func (s *Store) RawQuery(sql string, params ...interface{}) (ResultSet, error) {
-	rows, err := s.db.Query(sql, params...)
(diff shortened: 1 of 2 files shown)


erizocosmico:
Tests added @Serabe @dpordomingo

PR #137, opened 2017-04-25
Like operator

jorbs:


operators.go
```diff
@@ -302,6 +310,11 @@ type (
 		col    string
 		values []interface{}
 	}
+
+	like struct {
```
erizocosmico: there's no need for a new struct, actually, there is `colOp` for that https://github.com/src-d/go-kallax/blob/master/operators.go#L286

operators.go
```diff
@@ -62,6 +62,14 @@ func Neq(col SchemaField, value interface{}) Condition {
 	}
 }
 
+// Like returns a condition that will be true when `col` matches the given value.
+// See https://www.postgresql.org/docs/9.6/static/functions-matching.html.
+func Like(col SchemaField, value string) Condition {
+	return func(schema Schema) squirrel.Sqlizer {
+		return like{col.QualifiedName(schema), value}
```
erizocosmico: Use `colOp` instead of custom type https://github.com/src-d/go-kallax/blob/master/operators.go#L106

tests/query_test.go
```diff
@@ -87,6 +87,16 @@ func (s *QuerySuite) TestQuery() {
 	s.Panics(func() {
 		s.Equal("bar", store.MustFindOne(queryErr).Foo)
 	})
+
+	foobar := newQueryFixture("foo bar")
```
jorbs: Initially, I tried to run `operators_test.go` inside of my forked repo folder, but I always get `can't load package: package github.com/jorbs/go-kallax: code in directory ~/go/src/github.com/jorbs/go-kallax expects import "gopkg.in/src-d/go-kallax.v1"`. I'm new on Go lang, but I think the packages are fixed to the original repo (github.com/src-d/go-kallax).

erizocosmico:
Thanks! LGTM.

Can you remove the commit [`3d89d2b`](https://github.com/src-d/go-kallax/pull/137/commits/3d89d2b51c69fc2d5a8c831941e2abf1ccd052ca), both changes show up right now

tests/query_test.go
erizocosmico: Yes, they are fixed to `gopkg.in/src-d/go-kallax.v1`. What you can do is clone the repo at `$GOPATH/src/gopkg.in/src-d/go-kallax.v1` and add your own remote there. Unfortunately, until real package management is a thing in Go, we have to stick with such methodologies.

jorbs: I got it. Thanks.

erizocosmico:
Don't worry, I already cherrypicked the commit and pushed it. Thanks!

PR #146, opened 2017-05-04
implement sql migrations

### LINKED ISSUE #145: Implement migrations
As discussed in https://github.com/src-d/go-kallax/issues/136

There will be a PR with this soon.

erizocosmico:
Closes #145 

### Design overview

There are three types of migrations involved in migrations:
* `lock.json`: contains the latest version of the SQL schema. Every new generation, takes it from there and diffs the current schema inferred from the models against it.
* `TS_name.up.sql`: an up script for a migration.
* `TS_name.down.sql`: a down script for a migration.

### Flow

* Lock file is read.
* Models are processed and converted to a SQL schema (tables and columns).
* Diff between lock schema and current schema produces the change set for the up migration.
* With the up migration, the inverse of the changes are made and this is the down migration.
* With the up and down migrations and the current schema, three files are generated:
  * New lock with the current schema.
  * New up script.
  * New down script.

### Caveats

* A new struct tag for fields is introduced: `sqltype`. There are some cases in which we can't know which SQL corresponds to a Go type, for example, in `sql.Valuer`s. In this case, the generation will fail and the user will have to use a custom type.
* There are some cases where tracking certain changes are hard for us to do automatically, and thus, require manual change. There is a type of change named `ManualChange` that generates invalid SQL syntax (with a human-readable message so the user knows why the manual change is required) so the user can't run the migrations until a proper migration is done manually.
* Renames can't be tracked in neither tables nor columns, so they are a DROP+CREATE instead.

### Missing features

* This only introduces the generation of migrations, not their execution! That will be done in a followup PR.
* Once this is merged and finalized I will make a followup PR with the documentation for this feature.

### DIFF
--- a/generator/cli/kallax/cmd.go
+++ b/generator/cli/kallax/cmd.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 
@@ -37,8 +36,28 @@ func main() {
 		{
 			Name:   "gen",
 			Usage:  "Generate kallax models",
-			Action: app.Action,
+			Action: generateModels,
 			Flags:  app.Flags,
+		},
+		{
+			Name:   "migrate",
+			Usage:  "Generate migrations for current kallax models",
+			Action: migrateAction,
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "out, o",
+					Usage: "Output directory of migrations",
+				},
+				cli.StringFlag{
+					Name:  "name, n",
+					Usage: "Descriptive name for the migration",
+					Value: "migration",
+				},
+				cli.StringSliceFlag{
+					Name:  "input, i",
+					Usage: "List of directories to scan models from. You can use this flag as many times as you want.",
+				},
+			},
 		},
 	}
 
@@ -50,7 +69,12 @@ func generateModels(c *cli.Context) error {
 	output := c.String("output")
 	excluded := c.StringSlice("exclude")
 
-	if !isDirectory(input) {
+	ok, err := isDirectory(input)
+	if err != nil {
+		return fmt.Errorf("kallax: can't check input directory: %s", err)
+	}
+
+	if !ok {
 		return fmt.Errorf("kallax: Input path should be a directory %s", input)
 	}
 
@@ -69,11 +93,55 @@ func generateModels(c *cli.Context) error {
 	return nil
 }
 
-func isDirectory(name string) bool {
-	info, err := os.Stat(name)
-	if err != nil {
-		log.Fatal(err)
+func migrateAction(c *cli.Context) error {
+	dirs := c.StringSlice("input")
+	dir := c.String("out")
+	name := c.String("name")
+
+	var pkgs []*generator.Package
+	for _, dir := range dirs {
+		ok, err := isDirectory(dir)
+		if err != nil {
+			return fmt.Errorf("kallax: cannot check directory in `input`: %s", err)
+		}
+
+		if !ok {
+			return fmt.Errorf("kallax: `input` must be a valid directory")
+		}
+
+		p := generator.NewProcessor(dir, nil)
+		p.Silent()
+		pkg, err := p.Do()
+		if err != nil {
+			return err
+		}
+
+		pkgs = append(pkgs, pkg)
 	}
 
-	return info.IsDir()
+	ok, err := isDirectory(dir)
+	if err != nil {
+		return fmt.Errorf("kallax: cannot check directory in `out`: %s", err)
+	}
+
+	if !ok {
+		return fmt.Errorf("kallax: `out` must be a valid directory")
+	}
+
+	g := generator.NewMigrationGenerator(name, dir)
+	migration, err := g.Build(pkgs...)
+	if err != nil {
+		return err
+	}
+
+	return g.Generate(migration)
+}
+
+func isDirectory(name string) (bool, error) {
+	info, err := os.Stat(name)
+	if err != nil {
+		return false, err
+	}
+
+	return info.IsDir(), nil
 }
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -3,9 +3,18 @@
 package generator // import "gopkg.in/src-d/go-kallax.v1/generator"
 
 import (
+	"bytes"
+	"encoding"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime/debug"
+	"strings"
+	"time"
+
+	"github.com/fatih/color"
 )
 
 // Generator is in charge of generating files for packages.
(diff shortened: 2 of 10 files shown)


generator/migration.go
```diff
@@ -0,0 +1,670 @@
...
+// NewMigration creates a new migration from the old and the new schema.
+func NewMigration(old, new *ModelSchema) *Migration {
+	var migration = &Migration{}
+	migration.Up = SchemaDiff(old, new)
+	migration.Down = ReverseChangeSet(migration.Up, old)
+	migration.Lock = new
+	return migration
+}
+
+// ModelSchema represents a schema of all the models in the database.
```
Serabe: Then maybe it should be named DbSchema or ModelsSchema

generator/cli/kallax/cmd.go
```diff
@@ -69,10 +88,44 @@ func generateModels(c *cli.Context) error {
...
+	}
+
+	return g.Generate(migration)
+}
+
 func isDirectory(name string) bool {
 	info, err := os.Stat(name)
 	if err != nil {
-		log.Fatal(err)
+		return false
```
ajnavarro: There are some cases that the path will be a directory but `os.Stat()` will return an error, per example in permission errors.

generator/generator.go
```diff
@@ -49,3 +56,119 @@ func (g *Generator) writeFile(pkg *Package) (err error) {
...
+	}
+
+	return nil
+}
+
+func slugify(str string) string {
+	var buf bytes.Buffer
+	for _, r := range strings.ToLower(str) {
+		if ('a' <= r && r <= 'z') || ('0' <= r && r <= '9') {
+			buf.WriteRune(r)
```
ajnavarro: `WriteRune` returns an error.

generator/migration.go
```diff
@@ -0,0 +1,670 @@
...
+	SmallIntColumn    ColumnType = "smallint"
+	IntegerColumn     ColumnType = "integer"
+	BigIntColumn      ColumnType = "bigint"
+	RealColumn        ColumnType = "real"
+	DoubleColumn      ColumnType = "double"
+	SmallSerialColumn ColumnType = "smallserial"
+	SerialColumn      ColumnType = "serial"
+	BigSerialColumn   ColumnType = "bigserial"
+	TimestamptzColumn ColumnType = "timestamptz"
+	TextColumn        ColumnType = "text"
```
ajnavarro: For short strings could be better use a `varchar(n)` instead of `text`, How will we able to choose this?

generator/migration.go
```diff
@@ -0,0 +1,670 @@
...
+type AddColumn struct {
+	// Column schema.
+	Column *ColumnSchema
+	// Table to add the column to.
+	Table string
+}
+
+func (c *AddColumn) MarshalText() ([]byte, error) {
+	var buf bytes.Buffer
+	buf.WriteString(fmt.Sprintf("ALTER TABLE %s ADD COLUMN ", c.Table))
```
ajnavarro: `WriteString` returns error.

ajnavarro: I'll suggest:
```go
return []byte(fmt.Sprintf("ALTER TABLE %s ADD COLUMN %s ;\n", c.Table,c.Column)), nil
```

generator/migration.go
```diff
@@ -0,0 +1,670 @@
...
+func ReverseChangeSet(cs ChangeSet, old *ModelSchema) ChangeSet {
+	var result = make(ChangeSet, len(cs))
+	for i, c := range cs {
+		result[i] = reverseChange(c, old)
+	}
+	return result
+}
+
+// reverseChange generates the inverse of a change.
+func reverseChange(c Change, old *ModelSchema) Change {
```
ajnavarro: Instead of do this, could be better create a method called `Reverse()` into the `Change` interface. Then each `Change` implementation can implement his inverse. Also, doing this, we can delete this `panic("unreachable")` that I hate so much.

generator/migration.go
```diff
@@ -0,0 +1,670 @@
...
+	return nil, nil
+}
+
+var typeMappings = map[string]ColumnType{
+	"gopkg.in/src-d/go-kallax.v1.ULID":      UUIDColumn,
+	"gopkg.in/src-d/go-kallax.v1.UUID":      UUIDColumn,
+	"gopkg.in/src-d/go-kallax.v1.NumericID": BigIntColumn,
+	"github.com/satori/go.uuid.UUID":        UUIDColumn,
+	"string":        TextColumn,
+	"rune":          TextColumn,
```
ajnavarro: wow, too much for a rune, right?

generator/migration.go
```diff
@@ -0,0 +1,670 @@
...
+	"uint32":        BigIntColumn,
+	"int32":         IntegerColumn,
+	"uint":          NumericColumn(20),
+	"int":           BigIntColumn,
+	"int64":         BigIntColumn,
+	"uint64":        NumericColumn(20),
+	"float32":       RealColumn,
+	"float64":       DoubleColumn,
+	"bool":          BooleanColumn,
+	"url.URL":       TextColumn,
```
ajnavarro: we will need a way to set size of a string field for use `TextColumn` or `BarcharColumn(n)`

qq
```diff
@@ -0,0 +1,1764 @@
+[33mcommit c6fa08d87ca144357489fff474bc1f7020414b8e[m
```
ajnavarro: ?

generator/migration.go
erizocosmico: Yes, that name has been bugging me since the beginning. `DBSchema` sounds good.

generator/generator.go
erizocosmico: But it's always nil https://golang.org/src/bytes/buffer.go?s=8208:8261#L237

All the writes to `bytes.Buffer` return nil

generator/cli/kallax/cmd.go
erizocosmico: ok, I can put back the log

generator/migration.go
erizocosmico: same as `WriteRune`

generator/migration.go
erizocosmico: Sounds good

generator/migration.go
erizocosmico: Yes, I'm gonna put a `char(1)`

generator/migration.go
erizocosmico: it's already there, the struct tag `sqltype`

qq
erizocosmico: that was a letfover, will delete asap

generator/generator.go
ajnavarro: Good to know!

generator/migration.go
erizocosmico: Using `sqltype` struct tag

erizocosmico:
Done @Serabe @ajnavarro

Serabe:
Thank you, @erizocosmico

PR #174, opened 2017-06-13
fix relationshipRecords for non-ptr relationships

erizocosmico:
If you have N relationships, only the last one was taken into account. GGWP.

### DIFF
--- a/generator/templates/model.tgo
+++ b/generator/templates/model.tgo
@@ -123,12 +123,12 @@ func (s *{{.StoreName}}) relationshipRecords(record *{{.Name}}) []kallax.RecordW
         var records []kallax.RecordWithSchema
         {{range .NonInverses}}
         {{if .IsOneToManyRelationship}}
-        for _, rec := range record.{{.Name}} {
-                rec.ClearVirtualColumns()
-                rec.AddVirtualColumn("{{.ForeignKey}}", record.GetID())
+        for i := range record.{{.Name}} {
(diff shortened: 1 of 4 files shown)


roobre:
May I should note that the current release (without this patch) does not update ID fields of models inserted within one-level of recursion when the relationsip is not a slice of pointers.

This is, if `A` has many `B`s, after doing `astore.Save(&a)`, `a.bs[i].ID` is still 0. I'm not sure if this patch fixes that behaviour too.

roobre:
Also, (I apologize for the doublepost), and this is a little weird, if we have the following scenario:

```
type A struct {
    kallax.Model
    ID int64 `pk:"autoincr"`

    Bs []*B // Note the pointer
}


type B struct {
    kallax.Model
    ID int64 `pk:"autoincr"`

    Cs []C // Note this is not a pointer
}


type C struct {
    kallax.Model
    ID int64 `pk:"autoincr"`
}
```

Doing

```
astore.Save(&a) // This works, inserts a and its Bs, and sets `a.bs[i].ID` in memory
// for each a.bs
bstore.Save(&a.bs[i]) // cs are NOT saved
```

Does NOT trigger the insertion of any `C`, despite `a.bs[i].ID` being correctly set.

erizocosmico:
Ok, everything should be fixed now. I'm merging this and releasing. If it's not, please raise an issue.

Issue #114, opened 2017-03-28
Many to many relationships


erizocosmico:

RobbieMcKinstry:
Hi! :)

Is this likely to be addressed in a future release? Is there anything blocking this?

I don't think I'd be able to contribute to this issue specifically but I might be able to tackle a blocker.

erizocosmico:
I'm planning on starting to work on that as soon as I can, but there's not any release date yet. Sadly, the only blocker is lack of time :(

amattn:
Any recent progress on this issue?  Anything specific the community may be able to help out on?

PR #183, opened 2017-06-19
add manythrough support in queries and batcher, add through props to field

erizocosmico:
This PR is the first of several that will be merged into `feature/m2m`, which will be merged when full support for many to many relationships is completely finished.

### What is introduced in this PR?

* Accessing of several properties in `through` struct tag of a field via methods of `Field` type.
* Support both on queries and in the `batchRunner` for retrieving N:M relationships. 1:N using a through _should_ work but that is yet to be tested.

### Notable changes and breaking changes

* `Identifier` now has a `newPtr` method used to get a new fresh empty pointer of that identifier type (internal, just used in the batcher). This restricts the extension of the `Identifier` interface, though, which is probably not a big deal.
* `Scan` now accepts N extra pointers. Again, it's internal and not really exposed to the public, so we should be good.
* `ManyToMany` has been renamed to `Through`, as 1:N with an intermediate table also need to be processed as a `Through` kind of relationship.
* `ForeignKeys` has now a slice of foreign keys, instead of just one. Only generated code uses it, but it will be broken until regeneration.

### What's left?

* [ ] Schema generation
* [ ] Go boilerplate generation
* [ ] Inserts + updates with through relationships
* [ ] Test it works with 1:N with intermediate tables

At least, all the internal mumbo jumbo should be done with this PR.

### Review

I suggest you take a look at it per-commit, as they are very clear and separate functionalities. (Should have made 2 PRs actually, but /shrug)

### DIFF
--- a/batcher.go
+++ b/batcher.go
@@ -14,6 +14,7 @@ type batchQueryRunner struct {
 	q             Query
 	oneToOneRels  []Relationship
 	oneToManyRels []Relationship
+	throughRels   []Relationship
 	db            squirrel.DBProxy
 	builder       squirrel.SelectBuilder
 	total         int
@@ -29,6 +30,7 @@ func newBatchQueryRunner(schema Schema, db squirrel.DBProxy, q Query) *batchQuer
 	var (
 		oneToOneRels  []Relationship
 		oneToManyRels []Relationship
+		throughRels   []Relationship
 	)
 
 	for _, rel := range q.getRelationships() {
@@ -37,6 +39,8 @@ func newBatchQueryRunner(schema Schema, db squirrel.DBProxy, q Query) *batchQuer
 			oneToOneRels = append(oneToOneRels, rel)
 		case OneToMany:
 			oneToManyRels = append(oneToManyRels, rel)
+		case Through:
+			throughRels = append(throughRels, rel)
 		}
 	}
 
@@ -46,6 +50,7 @@ func newBatchQueryRunner(schema Schema, db squirrel.DBProxy, q Query) *batchQuer
 		q:             q,
 		oneToOneRels:  oneToOneRels,
 		oneToManyRels: oneToManyRels,
+		throughRels:   throughRels,
 		db:            db,
 		builder:       builder,
 	}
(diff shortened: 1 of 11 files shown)


erizocosmico:
@Serabe @dpordomingo 

There is a problem with the automatic insert+update of relationships with a table in between: it's a record that may need to be initialized in some way (imagine a database that used `UUID`s as IDs instead of numerical ones).

For example, imagine the following scenario:

```go
type User struct {
  kallax.Model
  ID uuid.UUID
}

type Post struct {
  kallax.Model
  ID uuid.UUID
}

type UserPost struct {
  kallax.Model
  ID uuid.UUID
  User *User
  Post *Post
}
```

In this case, the user will need to set the ID in the constructor, and this is necessary before inserting this record in the database. The problem there is that constructors can have arbitrary parameters and we're not magicians so we don't know how to call them behind the scenes. There's only so much magic we can do.

So, we can either think of a solution for this that is nice for the user (which I can't really think of right now) or one the following:

- Records for relationships with a `through` are read only and insertions need to be manual, but updates can be automatic.
- Records for relationships with a `through` are read only and insertions/updates **both** need to be manual.
- Create methods like `AddX`, `RemoveX`, `SaveX` for each through relationship in which you have to pass the already initialized through record.

Note that updates are not a problem, since they don't need the `through` record to be saved. The problem is only during insert.

Any thoughts?

Serabe:
I'm adding this to backlog, to keep track in the future.

roobre:
In the upcoming days, I'm merging this into a wip/feature branch in order to work more confortably from it. Please, consider discussing any issues on the open issue related to this (#114).

After merging this PR, destination branch will be rebased to master (I'm praying to the flying spaguetti monster, cthulu, or whoever can help me) and another PR will be created when the feature is complete (or, at least, usable) in case someone wants to review it.

iorlas:
Sooo... what's the status? Are you guys gone full go-mysql-server or what?

nadiamoe:
@iorlas Thanks for your interest. I cannot speak for src-d, but personally I lack the time needed to push this feature forward.

Issue #157, opened 2017-05-15
Empty file generated on windows


Alexendoo:
`go version go1.8 windows/amd64`

Using the test case:

```go
package main

import "gopkg.in/src-d/go-kallax.v1"

//go:generate kallax gen

type Foo struct {
        kallax.Model `table:"foos"`
        ID int64 `pk:""`
}
```

produces the full expected output on linux, but on windows only provides:

```go
// IMPORTANT! This is auto generated code by https://github.com/src-d/go-kallax
// Please, do not touch the code below, and if you do, do it under your own
// risk. Take into account that all the code you write here will be completely
// erased from earth the next time you generate the kallax models.
package main

import (
        "fmt"

        "gopkg.in/src-d/go-kallax.v1/types"
)

var _ types.SQLType
var _ fmt.Formatter

type schema struct {
}

var Schema = &schema{}
```

erizocosmico:
What version of kallax are you using? You can find out by doing `kallax version`
On Mon, 15 May 2017 at 13:04, Alex Macleod <notifications@github.com<mailto:notifications@github.com>> wrote:

go version go1.8 windows/amd64

Using the test case:

package main

import "gopkg.in/src-d/go-kallax.v1<http://gopkg.in/src-d/go-kallax.v1>"

//go:generate kallax gen

type Foo struct {
        kallax.Model `table:"foos"`
        ID int64 `pk:""`
}

produces the full expected output on linux, but on windows only provides:

// IMPORTANT! This is auto generated code by https://github.com/src-d/go-kallax
// Please, do not touch the code below, and if you do, do it under your own
// risk. Take into account that all the code you write here will be completely
// erased from earth the next time you generate the kallax models.
package main

import (
        "fmt"

        "gopkg.in/src-d/go-kallax.v1/types<http://gopkg.in/src-d/go-kallax.v1/types>"
)

var _ types.SQLType
var _ fmt.Formatter

type schema struct {
}

var Schema = &schema{}

Alexendoo:
`kallax version 1.2.1`

m-kad:
Interesting find, when I have kallax files inside vendor/ folder, I get "empty" kallax.go too, but when I delete kallax from vendor/ I get normal kallax.go file. Maybe there's problem somewhere.
`go version go1.8.1 linux/amd64`
`kallax version 1.2.1`

Alexendoo:
Ah odd, that happens to me on linux also, but with windows it's empty with either vendored or unvendored

[...]

PR #185, opened 2017-06-20
add provided example for error in issue #157

erizocosmico:
If it passes the build in AppVeyor, means kallax is OK and further information needs to be provided.

### DIFF
--- a/tests/issue_157.go
+++ b/tests/issue_157.go
@@ -1,0 +1,8 @@
+package tests
+
+import kallax "gopkg.in/src-d/go-kallax.v1"
+
+type Foo struct {
+	kallax.Model `table:"foos"`
+	ID           int64 `pk:""`
+}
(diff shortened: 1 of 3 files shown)


erizocosmico:
The makefile already runs go generate
On Tue, 20 Jun 2017 at 20:11, David <notifications@github.com<mailto:notifications@github.com>> wrote:

@dpordomingo commented on this pull request.

I'm not sure if I understood the issue properly, but if I did, the the problem reported was that Kallax does not generate a proper kallax.go file in Windows.
If so, this test does not try to reproduce the error, because it does not generate the kallax.go but it is already included in the code.

I think that what would be needed is that the "testcase" starts with only one file created: tests/issue_157.go

  1.  runs go generate
  2.  asserts that NewFoo() is generated under the generated kallax.go

Issue #171, opened 2017-06-08
Generating constraints


qJkee:
Is it's possible to set constrains to fields,(for example 'unique') or I need to check that fields manually by defining methods such as `BeforeUpdate`?

erizocosmico:
Right now there's no way to do that, but if you want to tackle that in a PR, it will be welcome'd!

If you're not using the migrations provided by kallax you can put that directly in your schema. If you are using them, you still can edit the SQL generated by kallax before running it, anyway.

qJkee:
Hm, I think I can find time to look at your project and try to implement this feature =)

erizocosmico:
Thanks! Just to give you some guidance, that would need to be struct tags, like 

```go
type Foo struct {
    SomeField int `unique:"true"`
}
```

Niondir:
I could also need this feature. 

I guess the `lock.json` will also need some field `"Unique": true`

Niondir:
I created this pr: https://github.com/src-d/go-kallax/pull/204

No test cases yet!

erizocosmico:
@Niondir Thanks! yes, the lock would also need the unique field!

Issue #203, opened 2017-07-14
types/types_test TestNullable failes locally on windows


Niondir:
> Error Trace:	types_test.go:434
> 	Error:		Not equal: 
> 			expected: time.Time{sec:63635636271, nsec:0, loc:(*time.Location)(nil)}
> 			received: time.Time{sec:63635636271, nsec:0, loc:(*time.Location)(0xc04203b3e0)}
> 			
> 			Diff:
> 			--- Expected
> 			+++ Actual
> 			@@ -1,2 +1,2 @@
> 			-(time.Time) 2017-07-14 13:37:51 +0000 UTC
> 			+(time.Time) 2017-07-14 15:37:51 +0200 CEST
> 			 
> 	Messages:	time.Time

roobre:
I have no means to test or fix this, as I don't have windows machines. PRs are welcome.

smola:
This fails on macOS too:

```diff
--- Expected
+++ Actual
@@ -1,2 +1,2 @@
-(time.Time) 2018-05-17 17:07:56 +0000 UTC
+(time.Time) 2018-05-17 17:07:56 +0000 GMT
```

See a failing macOS build here: https://travis-ci.org/src-d/go-kallax/jobs/380310885#L1042 (requires https://github.com/src-d/go-kallax/pull/268)

smola:
When creating the date, I guess we have to force it to UTC (with `.UTC()` function?), since each operating system translates the `+0000` offset to a different timezone by default. I guess this could even happen on Linux in some setups.

PR #204, opened 2017-07-14
Add unique constraints to columns with `unique:"true"`

Niondir:
https://github.com/src-d/go-kallax/issues/171

It's now possible to defined unique constraints for columns. We should discuss the format. Currently it's as proposed in https://github.com/src-d/go-kallax/issues/171

No constraint:
```
`unique:""`
`unique:"false"`
```

With constraint:
```
`unique:"true"`
`unique:"maybe-the-constraint-name-in-future"`
```

Other ideas are:

```
type Foo struct {
    SomeField int `unique:""`
}
```

```
type Foo struct {
    SomeField int `constraint:"unique"`
}
```

Postgres supports the following formats due to https://www.postgresql.org/docs/9.5/static/ddl-constraints.html

```
-- Implemented with this PR
CREATE TABLE products (
    product_no integer UNIQUE,
    name text,
    price numeric
);

-- Not yet supported
CREATE TABLE products (
    product_no integer CONSTRAINT must_be_different UNIQUE,
    name text,
    price numeric
);

-- Not yet supported
CREATE TABLE example (
    a integer,
    b integer,
    c integer,
    UNIQUE (a, c)
);
```

```
CREATE TABLE example (
    a integer,
    b integer,
    c integer,
    UNIQUE (a, c)
);
```

### DIFF
--- a/generator/types.go
+++ b/generator/types.go
@@ -642,6 +642,7 @@ type Field struct {
 
 	primaryKey      string
 	isPrimaryKey    bool
+	isUnique        bool
 	isAutoincrement bool
 	columnName      string
 }
@@ -709,8 +710,22 @@ func NewField(n, t string, tag reflect.StructTag) *Field {
 		primaryKey:      pkName,
 		columnName:      columnName(n, tag),
 		isPrimaryKey:    isPrimaryKey,
+		isUnique:        uniqueProperty(tag),
 		isAutoincrement: autoincr,
 	}
+}
+
+func uniqueProperty(tag reflect.StructTag) bool {
+	val, ok := tag.Lookup("unique")
+	if !ok {
+		return false
+	}
+
+	if val == "" || strings.ToLower(val) == "false" {
+		return false
+	}
+
+	return true
 }
 
 // pkProperties returns the primary key properties from a struct tag.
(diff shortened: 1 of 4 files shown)


Niondir:
Maybe it's better to only have no constraint if no unique tag is given at all. Especially when we want to use the tag value as constraint name in future.

generator/types.go
```diff
@@ -709,10 +710,24 @@ func NewField(n, t string, tag reflect.StructTag) *Field {
 		primaryKey:      pkName,
 		columnName:      columnName(n, tag),
 		isPrimaryKey:    isPrimaryKey,
+		isUnique:        uniqueProperty(tag),
 		isAutoincrement: autoincr,
 	}
 }
 
+func uniqueProperty(tag reflect.StructTag) bool {
+	val, ok := tag.Lookup("unique")
```
erizocosmico: you can go with `tag.Get("unique") == "true"`, all other options would be ignored until we decide if it's worth implementing the custom name

types/types_test.go
```diff
@@ -431,8 +431,10 @@ func TestNullable(t *testing.T) {
 		if c.isPtr {
 			elem = elem.Elem()
 		}
-		s.Equal(c.nonNullInput, elem.Interface(), c.name)
-
+		// TODO: can be commented in if tests fail locally until fixed.
```
erizocosmico: leftover?

erizocosmico: Is it because of https://github.com/src-d/go-kallax/issues/203? Then I'd prefer removing the comments, since this is another issue and the CI passes anyway.

PR #210, opened 2017-07-28
Added reproduction case for relations being deleted on Save()

roobre:
Reproduction case for https://github.com/src-d/go-kallax/issues/211

### DIFF
--- a/tests/kallax.go
+++ b/tests/kallax.go
@@ -18,6 +18,413 @@ import (
 var _ types.SQLType
 var _ fmt.Formatter
 
+// NewBrand returns a new instance of Brand.
+func NewBrand(name string) (record *Brand) {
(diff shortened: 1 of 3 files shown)


roobre:
Looks like `Person.Save()` saves only first level relationships, just like the README says. However, second-level relationships are destroyed if they already exist.

The most inmediate fix would be to avoid this behaviour, making `Update()` ignore second-level relationships instead of deleting them.

As, in most cases, one would want to change just the relationship itself and not the whole depending object, a plain `Update()` function may be appreciated.

 In fact, I think the case where you want to edit an object and its relatives in the same operation is fairly common for insertions, but relatively rare for updates. So making this new "plain update" the default behaviour and delegating the current "full update" operation to a different method might be worth considering.

PR #227, opened 2017-08-21
Printing CLI errors to stderr

### LINKED ISSUE #225: All CLI errors are silent
I've noticed that all CLI utility errors are silent. For example, if you try to create a migration, but you did something wrong, there will be no output and no files created.

I tracked this down to how the CLI is run. You do:
```go
newApp().Run(os.Args)
```

But according to [the docs for the cli lib version you're using](https://godoc.org/gopkg.in/urfave/cli.v1#App.Run), errors returned by cli actions must implement `cli.ExitCoder` to be printed, but the errors returned by `errors.New` definitely don't.

To fix, you can either:
- Switch to using the (technically deprecated) `RunAndExitOnError` function
- Print the error returned by `Run` and exit with a non-zero code
- Implement `cli.ExitCoder` for the errors (or wrap just as they escape the action function)

Out of those, I'd probably pick the first, since you've pinned the urfave cli library at v1 anyway.

If you'd prefer, I can send a PR, but it's a pretty small change that you could probably make more quickly yourselves. :smile:

jBugman:
Fixes #225.

### DIFF
--- a/generator/cli/kallax/cmd.go
+++ b/generator/cli/kallax/cmd.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"gopkg.in/src-d/go-kallax.v1/generator/cli/kallax/cmd"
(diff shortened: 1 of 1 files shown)


roobre:
Could you please read #238 and consider opening this PR there?

Thanks.

PR #236, opened 2017-09-21
"Exists" method for store

josephbuchma:
Example usage:
```go
exists, err := NewUserStore(db).Exists(NewUserQuery().Where(kallax.Eq(Schema.User.Email, form.Email)))
```


roobre:
I'd say this functionality is already covered with `Count()`, which returns zero if no records are found. It looks a bit prettier tho, you have a point there.

josephbuchma:
I haven't looked at the code, but count is not efficient way of performing "exists" check.

roobre:
[AFAIK](https://wiki.postgresql.org/wiki/Index-only_scans#Is_.22count.28.2A.29.22_much_faster_now.3F), modern postgres versions use indices to perform simple counts. This should be quite fast (O(n), but in memory).

If I read your code correctly, you perform a query and call `rs.Next()`, which actually copies values to memory. Only the PK is copied tho (which I missed when I looked at it before), so your method is probably faster.

This seems good to me. If you want to add this PR to my fork (see #238), I'll happily merge it, given that at least two test cases (to check existence and non existence, preferably with non-trivial queries) and one or two lines at the README are added too.

roobre:
Now that I maintain this project actively, I'm interested on getting this merged. Please add some documentation and usage examples to `README.md`, as well as some tests to the suite. Then, unless anyone has anything to object, I'll merge it.

josephbuchma:
Hey @roobre , unfortunately for kallax I switched to sqlboiler. For now I don't have a time for finishing this, maybe later if it's not urgent. Otherwise example usage is provided in my first comment above, and testing this should be straight-forward, so you can get it merged.

PR #249, opened 2017-12-25
Fix relationship record retrieval

ligustah:
If a custom filter condition is supplied to any of the With{Type} methods the foreign key condition is not actually added to the query. While this doesn't break the retrieval it does pull a lot of unnecessary data into memory.

For a type A that has a 1:N relationship to type B the queries look like these:
```sql
SELECT __a.id, __a.c_id FROM a __a LIMIT 1 OFFSET 0;

SELECT __b_Other.id, __b_Other.attribute, __b_Other.a_id FROM b __b_Other WHERE __b_Other.attribute = $1;  -- current
SELECT __b_Other.id, __b_Other.attribute, __b_Other.a_id FROM b __b_Other WHERE (__b_Other.attribute = $1 AND __b_Other.a_id IN ($2)); --fixed
```

```golang
type A struct {
	kallax.Model
	ID    int64 `pk:"autoincr"`
	Other []B
}

type B struct {
	kallax.Model
	ID        int64 `pk:"autoincr"`
	Attribute int64
}
```

### DIFF
--- a/batcher.go
+++ b/batcher.go
@@ -175,7 +175,7 @@ func (r *batchQueryRunner) getRecordRelationships(ids []interface{}, rel Relatio
 
 	filter := In(fk, ids...)
 	if rel.Filter != nil {
-		And(rel.Filter, filter)
+		rel.Filter = And(rel.Filter, filter)
 	} else {
 		rel.Filter = filter
 	}


roobre:
Hello, thanks for the patch and PR. Could you post the go code which generates those queries? Just to make the issue clearer to me.

ligustah:
Sure thing:
```golang
package main

import (
	"database/sql"
	"log"

	kallax "gopkg.in/src-d/go-kallax.v1"
)

type A struct {
	kallax.Model
	ID    int64 `pk:"autoincr"`
	Other []B
	C     *C `fk:",inverse"`
}

type B struct {
	kallax.Model
	ID        int64 `pk:"autoincr"`
	Attribute int64
}

type C struct {
	kallax.Model
	ID       int64 `pk:"autoincr"`
	Property int64
}

func main() {
	db, err := sql.Open("postgres", "postgres://postgres:postgres@localhost/postgres?sslmode=disable")
	if err != nil {
		log.Fatal(err)
	}

	err = NewAStore(db).Transaction(func(aStore *AStore) error {
		var bStore BStore
		var cStore CStore
		kallax.StoreFrom(&bStore, aStore)
		kallax.StoreFrom(&cStore, aStore)

		c := &C{
			Property: 5,
		}

		err = cStore.Debug().Insert(c)
		if err != nil {
			return err
		}

		bs := []B{
			{Attribute: 0},
			{Attribute: 0},
			{Attribute: 1},
			{Attribute: 1},
			{Attribute: 2},
			{Attribute: 3},
		}

		return aStore.Debug().Insert(&A{
			Other: bs,
			C:     c,
		})
	})

	if err != nil {
		log.Fatal(err)
	}

	aStore := NewAStore(db).Debug()
	a, err := aStore.FindOne(
		NewAQuery().
			WithC().
			Where(kallax.Eq(Schema.C.Property, 5)),
	)
	if err != nil {
		log.Println(err)
	}

	log.Printf("%+v", a)

	a, err = aStore.FindOne(
		NewAQuery().
			WithOther(kallax.Eq(Schema.B.Attribute, 1)),
	)
	if err != nil {
		log.Println(err)
	}

	log.Printf("%+v", a)
}
```
That _should_ be it. I used it to test this one as well as #250

roobre:
Understood. Given the definition of `And`, it seems that someone slipped and forgot to reassign the result to the actual filter.

Looks good to me, thank you!

PR #255, opened 2018-01-20
Add a ToSql method to BaseQuery to expose query args

film42:
The BaseQuery.String method returns the prepared SQL but does not return
the query arguments, making it hard to test queries without mocking the
database. By adding a BaseQuery.ToSql method that returns the args as
well as an error returned while compiling, anyone can easily test
queries built by a helper function and verify the sql and args are
correct.

Here's an example of how I'd like to use this in my current project:

```golang
func Test_User_ByRequestFields_DateRange_UpdatedAt(t *testing.T) {
	time1 := time.Date(2018, time.January, 19, 13, 10, 23, 0, time.UTC)
	time2 := time.Date(2018, time.January, 19, 11, 10, 23, 0, time.UTC)
	time3 := time.Date(2018, time.January, 18, 10, 10, 23, 0, time.UTC)
	time4 := time.Date(2018, time.January, 18, 3, 10, 23, 0, time.UTC)

	request := &amigo.UserRequest{
		UpdatedAt: []*atlas.DateRange{
			newDateRange(time1.Unix(), time2.Unix()),
			newDateRange(time3.Unix(), time4.Unix()),
		},
	}

	testQuery, _ := User.ByRequestFields(nil, request)
	testQuerySQL, testQueryArgs, _ := testQuery.ToSql()
	expectedQuerySQL, expectedQueryArgs, _ := models.
		NewUserQuery().
		WithPromotion().
		WithClient().
		Where(kallax.In(models.Schema.User.IsDeleted, false, nil)).
		Where(kallax.Or(
			Between(models.Schema.User.UpdatedAt, "2018-01-19T13:10:23Z", "2018-01-19T11:10:23Z"),
			Between(models.Schema.User.UpdatedAt, "2018-01-18T10:10:23Z", "2018-01-18T03:10:23Z"))).
		ToSql()

	AssertEquals(t, "sql", expectedQuerySQL, testQuerySQL)
	AssertEquals(t, "sql args", expectedQueryArgs, testQueryArgs)
}
```
NOTE: Without verifying arguments, it's hard to say whether I'm actually searching for "users where is_deleted = false or nil" or if the dates are being formatted correctly for my between operator. 

Maybe I'm going about testing this incorrectly? I'm still pretty new to golang so it's very possible I missed something obvious in kallax to do this in a more idiomatic way.


roobre:
Hello and thanks for your effort on this PR.

I have a major question regarding this issue: What would this be used for in practice? I see you mention it is for testing, but I don't know if you mean kallax' own unit tests, or users debugging their code.

For the first case, kallax has a `query_test.go` which belongs to the same package as `query.go`, so we can write tests for it without exporting additional methods. For the second case, I just don't see the pint: Kallax has a lot of other features which can't be used without a database (mock, docker, test server, or whatever), so it doesn't make sense to me to support "printf debugging" for this particular piece. Users can easily turn on debugging on stores (`store.Debug()`) and check the whole queries there. I see no point in encouraging users to check SQL generated by kallax: they should assume it generates it correctly. That's why we have unit tests.

film42:
@roobre The code snippet is copied right out of my current project which lives in its own package. I have code that I use to turn a protobuf request into a large db query. The problem though is that there's no way for me to test that I'm setting the correct values in my query. I can check the SQL, but cannot check the arguments provided are correct without using a database driver. Since it's possible to get values from the `Condition` interface, I think `BaseQuery` implementing the `ToSqler` interface makes a lot of sense.

For example:

I can get access to the query args from a `kallax.In` instance:

```golang
condition := kallax.In(models.Schema.User.Username, WhereInStrings(request.Username)...)
toSqler := condition("example_schema_name")
sql, args, err := toSqler.ToSql()

// sql => SELECT ...
// args => []interface{}{"film42", "roobre", etc..}
```

But there's no way for me to access the `args` from the composed query:

```golang
query.Where(kallax.In(models.Schema.User.Username, WhereInStrings(request.Username)...))

sql := query.String()
// How can I verify I set my args correctly?
```

I could use something like [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock) and verify a select query is executed with specific arguments:

```golang
	testQuery, _ := User.ByRequestFields(nil, request)
	testQuerySQL := testQuery.String()
	expectedQuerySQL := models.
		NewUserQuery().
		WithPromotion().
		WithClient().
		Where(kallax.In(models.Schema.User.IsDeleted, false, nil)).
		Where(kallax.In(kallax.NewSchemaField("client_id"), 1, 3)).
		String()

	AssertEquals(t, "sql", expectedQuerySQL, testQuerySQL)

	db, mock, err := sqlmock.New()
	if err != nil {
		t.Fatalf("Error opening a stub database connection: %s", err)
	}
	defer db.Close()

	mock.ExpectPrepare("SELECT .*").ExpectQuery().WithArgs("film42", "roobre")
	store := models.NewUserStore(db)
	store.FindAll(testQuery)
	if err := mock.ExpectationsWereMet(); err != nil {
		t.Errorf("There were unfulfilled query expectations: %s", err)
	}
```

But this seems really overkill since the behavior to obtain values is possible (hence this PR), but it's not given the the user outside of the `kallax` package.

roobre:
I understand the you want to check the generated query, but I do not understand why, as it should be assumed kallax generates valid queries. If you want to debug them while you're developing your code, just set up a mock, or a postgres docker, or whatever you are experimenting with, and request a debug store. 

Or are you trying to check generated queries programatically?

film42:
Yes. I’m checking my generated queries programmatically. More specifically, I’m not trying to test kallax code to ensure it can generate correct queries. I know it does. I’m trying to test that my code can build queries using the correct values and columns programmatically, so it’s very important I check the values in my project’s tests.

roobre:
I'm still unsure about how this could help the majority of kallax userbase. In fact, I'd avoid that kind of checks if I were you. Programatically checking if your SQL is valid seems like an overkill to me: If you know how to check for its validity, you can probably generate valid sql on the first time.

Anyway, I'll drop another suggestion: What if you encapsulate your generated code on a transaction? You can add a custom logger to the transaction store, capture and parse the sql output, and if you don't like it, just issue a rollback. Of course this is acceptable if you can anticipate the query will succeed in most of the cases, otherwise you'll be wasting a lot of resources.

film42:
So for my case it’s really important to verify the sql and args I’m building because my request struct might have 1 of 25 fields populated. The query should only include “where” clauses for populated fields. The project I’m working on is a port of a ruby on rails service so having tests that can validate compatibility is critical. Really I don’t care about the actual SQL, I care that I’m building the query correctly with the correct arguments (i.e. I’m not making any typos).

I think your suggestions would work, but are very heavy work arounds for something I expected the `BaseQuery.String` to do by default (include the query args), since other types in kallax provide a way to access them.

I totally see your point about not wanting to export more and more methods. Change for the sake of change doesn’t make sense. But in this case I reached for a tool in the kallax toolbox and found one was missing (hence this PR). BTW, when do you feel it’s appropriate to add an exported method to a struct?

Anyway, I really love kallax and I’m just trying to make it better. Thanks for working with me on this! ❤️

film42:
Btw, @roobre . I put a gist together to show how I'm going about writing code and testing. Hopefully it provides a little more context: https://gist.github.com/film42/e38dc19bcd4af30b687fa0212b568f64

You'll notice I'm not actually trying to manipulate SQL, but attempting to correctly validate:
1. that my queries were built correctly with kallax methods.
2. that my code did not make typos by, for example, accidentally searching for "username" with "last_name" or something.

You'll see I found a way to work around (2), but the workaround seems a little crazy considering the query instance is holding the args but does not provide a way to get to them without mocking the database.

roobre:
Oh, so I'm finally starting to understand what you want here: You want Kallax to expose the generated sql in order to check your own bussiness logic/app code. This is a very bad approach.

You should always test your application end to end, if possible: 

1. Write some scripts which lifts a Postgres docker, or a test server, or whatever you're comfortable with. 
2. Add a setup for your test suite where you populate this database from the ground up, including schemas if possible (you can use kallax for that), and insert some sample data in a state you perfectly know and control. 
3. Write tests for your operations (add friend, delete friend, etc.) and checking if the data has been modified according to your expectations (e.g. User A and user B are friends, call `userA.Unfriend(userB)`, then assert userA is no longer friend of userB.
4. Wipe the testing environment completely (either by stopping the container, or with a script if you have it on your own machine) so it is on a blank state for the next test suite run.

To sum up, do the real thing on a testing environment and check against the results. Testing again the generated SQL is going to catch a little amount of errors, and require a lot of maintenance if your code (or kallax') is modified.

film42:
@roobre That's how I usually go about testing a web service, or anything with writes, but in this case, I'm creating a translation layer, where I need to testing turning a protobuf request into a complex select query. Complex enough that adding seed data to the db and verify results becomes extremely unmaintainable. Anyway, I don't think we need to align our testing philosophies since this PR allows both philosophies to happily coexist.

quixoten:
@roobre, maybe I'm incorrect in my analysis, but I think @film42 has provided reasonable justifications in response to each of your concerns. Looking at the change itself, I might be tempted to classify it as benign, but certainly it brings no harm to current users. It might not even be a stretch to consider this change within the realm of functionality one could reasonably consider a natural part of the library. Hopefully you can at least agree that the aim here is to improve kallax overall. I understand that you disagree with the particular use case that generated this PR, but I hope that doesn't prevent the addition from being considered in and of itself. There is at least one user who would greatly benefit from this change, and no users who it would unarguably harm. Are you open to reconsidering your position on this PR, or is it your recommendation that @film42, if he needs this functionality, must fork this project to attain it?

roobre:
> @roobre, maybe I'm incorrect in my analysis, but I think @film42 has provided reasonable justifications in response to each of your concerns. Looking at the change itself, I might be tempted to classify it as benign, but certainly it brings no harm to current users. It might not even be a stretch to consider this change within the realm of functionality one could reasonably consider a natural part of the library. Hopefully you can at least agree that the aim here is to improve kallax overall. I understand that you disagree with the particular use case that generated this PR, but I hope that doesn't prevent the addition from being considered in and of itself. There is at least one user who would greatly benefit from this change, and no users who it would unarguably harm. Are you open to reconsidering your position on this PR, or is it your recommendation that @film42, if he needs this functionality, must fork this project to attain it?

I'm currently in process of scalating this request to the original authors. Expanding the API surface has a friction cost, and even if it could be helpful for one user, it's certainly out of scope for the other 99%.
I appreciate @film42's efforts and PR, but I don't like this kind messages where third persons put pressure on issues which don't affect them by stating obvious things and asking rethorical questions. Please avoid this kind of behaviour in the future.

As I've said, the original authors have been contacted, in one or two days I'll have an answer for @film42 regarding this patch.

roobre:
Status update: After some discussion, we have concluded this feature is harmless for most users, and it can be useful for some corner scenarios. As the fingerprint is very low, this will be merged soon.

Some small changes might be required tho, @film42. Stay tuned.

film42:
I think @quixoten brought up some good points, but I'm glad to hear it's being considered. Happy to make any changes. Just let me know and I'll have changes ready ASAP.

PR #266, opened 2018-04-26
Change kallax.NumericID as PK from serial to bigserial

kaasasolut:
`serial` and `bigint` do not match. `serial` is 4 bytes long, while `bigint` is 8 bytes long.
See [the documentation of the current version](https://www.postgresql.org/docs/current/static/datatype-numeric.html#DATATYPE-NUMERIC) and [of the oldest version I could find, 7.2 from 2002](https://www.postgresql.org/docs/7.2/static/datatype.html#DATATYPE-NUMERIC).

Because the type of `NumericID` is `int64`, I changed `serial` to `bigserial`.


roobre:
Looks nice to me. I'd need to do a quick test to see how this affects migration schemas generated with `serial` tho, to ensure it is not breaking.

PR #268, opened 2018-05-17
ci: test on multiple postgres versions and macOS

smola:
* Add tests for macOS on Travis.
* Test multiple PostgreSQL versions ([ci-tricks](https://github.com/smola/ci-tricks/) sets them up based on the `POSTGRESQL_VERSION` variable).
* Simplify some parts of `.travis.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,31 @@
 language: go
-
-go:
-  - 1.9.x
-  - 1.10.x
-  - tip
+go_import_path: gopkg.in/src-d/go-kallax.v1
+sudo: required
 
 matrix:
-    allow_failures:
-        - go: tip
-    fast_finish: true
-
-addons:
-  postgresql: "9.4"
+  allow_failures:
+    - go: tip
+  fast_finish: true
+  include:
+    - {os: linux,                    go: 1.9.x,  env: 'POSTGRESQL_VERSION=9.6'}
+    - {os: linux,                    go: 1.10.x, env: 'POSTGRESQL_VERSION=9.4'}
+    - {os: linux,                    go: 1.10.x, env: 'POSTGRESQL_VERSION=9.5'}
+    - {os: linux,                    go: 1.10.x, env: 'POSTGRESQL_VERSION=9.6'}
+    - {os: linux,                    go: tip,    env: 'POSTGRESQL_VERSION=9.6'}
+    - {os: osx, osx_image: xcode9.3, go: 1.10.x, env: 'POSTGRESQL_VERSION=9.6'}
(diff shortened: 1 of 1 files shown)


roobre:
I'm not happy with adding a third-party dependency to the CI script. In the event the file becomes unavailable, or it has some errors, it can easily break the whole build process in a way hard to detect.

smola:
@roobre Sure. I'd expect not to merge this until https://github.com/src-d/go-kallax/issues/203 is fixed, which seems the only bug in macOS.

With respect to the 3rd-party dependency, you're doing the same with all other dependencies in the code, which are not vendored ;-)

Note that we are using the same dependency on [src-d/ci](https://github.com/src-d/ci/). So if it disappeared, we'd either bring it back or fix the whole thing for the other projects relying on this.

roobre:
> With respect to the 3rd-party dependency, you're doing the same with all other dependencies in the code, which are not vendored ;-)

Most (all?) other deps are large repos, which of course can disappear but it is not quite likely.

> Note that we are using the same dependency on src-d/ci. So if it disappeared, we'd either bring it back or fix the whole thing for the other projects relying on this.

Oh, if you guys take care of it then it's a different story. I didn't notice it at first glance, my bad. But just out of curiosity, are we using that only for the different versions of postgres? Is it really worth it?

Feel free to ping me when a fix arrives, then I'll happily merge it :)

smola:
> Most (all?) other deps are large repos, which of course can disappear but it is not quite likely.

Fair enough.

> Oh, if you guys take care of it then it's a different story.

Yes, me or someone else will take care if needed.

> But just out of curiosity, are we using that only for the different versions of postgres? Is it really worth it?

It actually started out of a set of Bash scripts to setup PostgreSQL and RabbitMQ. Then it became clear that these required dedicated testing on [rather extensive build matrix](https://travis-ci.org/smola/ci-tricks). So _repackaging_ this as Go seemed to be easier for me to run, test, etc.

I expect to add some more polyfill-like features for Travis and Appveyor, such as setting up other services and some runtimes such as Python.

You can add some inline scripts in `.travis.yml` and `.appveyor.yml` to achieve the same, but this is pretty fragile, because some steps need to change of different combinations of PostgreSQL, `os`, `dist/osx_image`, so testing this separately allows us to offload this setup effort from each project.

mcuadros:
@smola ?

mcuadros:
@smola ping

smola:
Closing. PostgreSQL install is not so consistent in new Travis images and I could not keep up with ci-tricks updates.

Issue #263, opened 2018-04-25
Memory/resource leak in postgresql process (probably by prepared statements)


enomado:
Hi guys. I'm doing some performance-driven development and just and notice that my computer go to swap. All memory was consumed by postgresql processes. 

It was just simple api called by `wrk tool` selecting table with something like

```
	store := md.NewCategoryStore(db)
	q := md.NewCategoryQuery()
	cats, err := store.FindAll(q)
```

`db.Exec("ROLLBACK;")` doesnt help
`db.Begin() ... tx.Rollback ` help but i didnt understand how.

I'm used to feel that nothing can leak in postgresql session and started thinking what it could be. Than i've noticed power using of prepared statements and gotcha, `db.Exec("DEALLOCATE ALL")` prevent leaking.

roobre:
Hello,

I've been investigating a bit about this, and in order to check if kallax is responsible for this, I would need you to check if the leak is produced with squirrel's statement cacher disabled.

You can do it like this:

```go
store := md.NewCategoryStore(db)
store = store.DisableCacher() // !! DisableCacher returns a store with the cacher disabled
q := md.NewCategoryQuery()
cats, err := store.FindAll(q)
```

enomado:
Oh I see. https://github.com/src-d/go-kallax/commit/75a61190be346416056d88c21e932eea39f62ae7
My version suddenly don't have this method. 

I can't re-generate models for the new version for now. From the beginning it was done by
`~/go/bin/kallax` but now it prints:
`could not import ./models (can't find import: "models")`

I lose my passion about `go` to investigate more time to figure out how to fix it. Probably best thing I can do is to share my code? https://gist.github.com/enomado/eeb101b7c04f2af822386be7803ff2e2
It was originally a playground of few orms... It was nothing more than simple query flood **without a commit**. Its a rare case anyway.

denchenkold:
If you are doing `store := md.NewCategoryStore(db)` on each api call, then it is expected behaviour, because squirrel stmt cacher does not find any prepared statements in local cache and executes `PREPARE ...` on database over and over. Thus, all these prepared statements live entire lifetime of db connection.

PR #269, opened 2018-05-28
Fix the sqlboiler benchmark

aarondl:
Did not update the readme numbers since I don't have the same hardware and
don't have the other ORMs set up properly.

```
BenchmarkKallaxInsertWithRelationships-8      	    1000	   1536213 ns/op	   10991 B/op	     225 allocs/op
BenchmarkSQLBoilerInsertWithRelationships-8   	    1000	   1646870 ns/op	    7328 B/op	     215 allocs/op
```

This patch significantly speeds up the InsertWithRelationships benchmark. Typically transactions should be being used in calls like this. The API that SQLBoiler has makes it a concious choice when to use them, props to Kallax for having these happen automatically when appropriate.

Also fixes an error where the benchmarks no longer run (go vet now runs automatically as of Go 1.10).

### DIFF
--- a/README.md
+++ b/README.md
@@ -829,7 +829,7 @@ BenchmarkKallaxUpdate-4                       	     300	   4179176 ns/op	     65
 BenchmarkKallaxUpdateWithRelationships-4      	     200	   5662703 ns/op	    6642 B/op	     175 allocs/op
 
 BenchmarkKallaxInsertWithRelationships-4      	     200	   5648433 ns/op	   10221 B/op	     218 allocs/op
-BenchmarkSQLBoilerInsertWithRelationships-4   	     100	  17543073 ns/op	    5898 B/op	     185 allocs/op
+BenchmarkSQLBoilerInsertWithRelationships-4   	     XXX	   XXXXXXX ns/op	    XXXX B/op	     XXX allocs/op
 BenchmarkRawSQLInsertWithRelationships-4      	     200	   5427503 ns/op	    4516 B/op	     127 allocs/op
 BenchmarkGORMInsertWithRelationships-4        	     200	   6196277 ns/op	   35080 B/op	     610 allocs/op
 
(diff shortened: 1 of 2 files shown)


roobre:
Hello,

Looks fine to me and I certainly can merge this right now, but benchmark results on `README.md` probably wouldn't get updated until the next release (which honestly can take a while). In order to be fair, I'm perfectly OK if you want to add a note to the benchmark results on `README.md` (like `*This results are outdated as of 2018-05-28 and will get updated soon`, or something like that).

aarondl:
Thanks @roobre. I did take the liberty! I hope the changes I've made are acceptable and thanks for being fair about it and allowing me to make the disclaimer.

It was a lot of fun profiling kallax and sqlboiler (did some flame graphs) and also seeing the different approaches we both took in the APIs.

I'd also like to share something I found interesting about the way kallax works too. If you look at it's use of the squirrel cache, that cache is never cleared. Prepared statements use memory on the database and the client and are meant to have `Close()` called on them when they're finished being used, and in Kallax this grows unbounded depending on how many queries you have in your app. I also found in my benchmarking that it barely affects performance at all, and is almost surely a detriment if you're only executing a query once or infrequently. Food for thought anyway.

Thanks :)

roobre:
> I hope the changes I've made are acceptable and thanks for being fair about it and allowing me to make the disclaimer.

They look indeed good to me. Of course, it's the least I can do :D

> [...] that cache is never cleared. Prepared statements use memory on the database and the client and are meant to have `Close()` called on them when they're finished being used, and in Kallax this grows unbounded depending on how many queries you have in your app.

I was aware of memory being leaked (while investigating #263), but not about the exact place. Thanks for sharing your findings!

>[Statement caching] barely affects performance at all, and is almost surely a detriment if you're only executing a query once or infrequently.

That's certainly something I did not know. As I only miantain kallax recently, I don't know why prepared statements where chosen on the first place. Perhaps for the free sql sanitizing? What I do know is `squirrel` was a temporary dependency, and kallax was meant to generate its own sql in the future. That plan was unfortunately abandoned, since the original authors shifted their efforts away from kallax.

A fix for that memory leak is something that indeed needs addressing. I will try to get some free time and actually work on a fix.

Anyway, I'm merging this to get the README updated. I'll hopefully rerun the benchmarks and get them a full refresh soon. Thanks for your effort!

aarondl:
Thanks for the merge. A couple more notes on caching/performance/prepared statements:

I benchmarked 10,000 inserts with a for loop, prepared statements, and a transaction. They took 8.6s, 8.2s and 800ms respectively. I also have a hunch that the sql library itself uses prepared statements for every single query by default for the sql sanitizing, so creating one manually doesn't affect performance much. It seems a bit improbable so I just re-checked and those results are consistent.

There's a way to turn off the cache so you can test easily https://godoc.org/gopkg.in/src-d/go-kallax.v1#Store.DisableCacher (not sure if you know since you aren't one of the original authors) which affects this: https://github.com/src-d/go-kallax/blob/v1.3.4/store.go#L138

PR #270, opened 2018-06-06
fix NewULID for systems with low time resolution

smola:
NewULID uses a pool of rand.Source using time.Now() as seed. In some systems with high concurrency and lower time resolution, this leads to multiple sources with the same seed, making NewULID return equal values on subsequent calls.
    
This commit changes the seed to add a random number from the global shared rand.Source (which is thread-safe).

### DIFF
--- a/model.go
+++ b/model.go
@@ -229,7 +229,8 @@ type Record interface {
 
 var randPool = &sync.Pool{
 	New: func() interface{} {
-		return rand.NewSource(time.Now().UnixNano())
+		seed := time.Now().UnixNano() + rand.Int63()
+		return rand.NewSource(seed)
 	},
 }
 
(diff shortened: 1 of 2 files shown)


smola:
Bug confirmed on Windows:

```
=== RUN   TestUULID_ThreeNewIDsAreDifferent
--- FAIL: TestUULID_ThreeNewIDsAreDifferent (0.02s)
	model_test.go:54: 
			Error Trace:	model_test.go:54
			Error:      	Not equal: 
			            	expected: 100000
			            	actual  : 72049
			Test:       	TestUULID_ThreeNewIDsAreDifferent
```

https://ci.appveyor.com/project/mcuadros/go-kallax/build/build-219.master

roobre:
I'll merge this straight away, but I want to note we should definitely use a cryptographic rng here. There are no reasons for not using cryptorngs on any scenario where randomness is needed and the code is not running baremetal (i.e. embedded/microprocessors).

PR #272, opened 2018-06-15
test: use DBHOST while running tests

smola:
This allows us to run tests against a PostgreSQL instance on a different host or a non-default port.


roobre:
Looks nice to me.

Issue #279, opened 2018-06-28
squirrel.DBProxyContext is undefined


faide:
when running a simple go build in the `gopkg.in/src-d/go-kallax.v1` directory (resulting from a go get) I get the following error:

```bash
# gopkg.in/src-d/go-kallax.v1
./store.go:120:12: undefined: squirrel.DBProxyContext
./store.go:121:12: undefined: squirrel.DBProxyContext
```

In store.go you import `github.com/Masterminds/squirrel` but looking at the `master` of squirrel I do not find any `DBProxyContext`.
If I checkout `gopkg.in/Masterminds/squirrel.v1` I find it.

I suppose you should import gopkg.in/Masterminds/squirrel.v1 and then in a second time investigate if the DBProxyContext is gone for real in the next squirrel release and find a way to stop using it in kallax.

faide:
After inspection it seems kallax uses the `DBProxyContext` but only uses the interface of `DBProxy` so a simple replacement could do the trick.

But, unfortunately it seems squirel Master also removed `Lt` and `Gt` from `expr.go` which kallax makes use of.

I'll change the ticket name to "squirrel incompatible changes"

faide:
After even more code sifting I infer that you are indeed using squirrel `v1` instead of `master` (master not being used by squirrel itself)

So in fact I think kallax should import `squirrel.v1` from `gopkg.in` instead of `github.com`

I'll propose a PR to do this

PR #280, opened 2018-06-28
Use gopkg.in/Masterminds/squirrel.v1 instead of github.com/Masterminds/squirrel

faide:
because the github version pulls from master and is not in synch with what kallax uses from squirrel.

See #279 for more reference


roobre:
Which go version are you using? According to squirrel's README:

> NOTE: as of Go 1.6, go get correctly clones the Github default branch (which is v1 in this repo).
> `import "github.com/Masterminds/squirrel"`

PR seems harmless and more explicit tho, so I'm inclined to merge it. Just want to know which is the root cause of the problem.

faide:
I am using go 1.10.3

Le 1 juil. 2018 2:30 PM, "Roberto Santalla" <notifications@github.com> a
écrit :

WnP:
I'm using `go version go1.11.1 linux/amd64` and have the same issue, this PR fix it.
Thanks @faide

Issue #205, opened 2017-07-17
With*() Query does not work with Count()


Niondir:
`store.Count(model.NewDeviceQuery().WithCustomer())` results in `pq: Spaltenverweis „id“ ist nicht eindeutig`

Is this intended, or should the With* part be ignored during `Count()`

erizocosmico:
Can you paste the output changing `store.Count` by `store.Debug().Count`? It's definitely not by design.

erizocosmico:
Just tried this and it worked for me. Maybe it's because it's trying to do `COUNT(identifier)`. Changed it to `COUNT(alias.*)` and now it should work. I'm gonna make a new release. Tell me if that solves it for you.

Note that adding `With*` to a query does not change the result of the `Count` since this is a left join and the result will always be the number of devices in your case no matter if they have a customer or not.

PR #291, opened 2018-08-08
Select star without schema alias in count

minond:
This change is to update the Count function to generate `SELECT COUNT(*) ...` instead of `SELECT COUNT(__pet.*) ...`

We are in the process of updating an application at work and we noticed that the existing system was able to handle certain requests much faster than our new one. Upon further inspection we narrowed it down to a fairly simple count query being generated by kallax.

When kallax generates a count query, it includes the schema's alias in the `COUNT` function ever since a fix for [this](https://github.com/src-d/go-kallax/issues/205) issue was [merged](https://github.com/src-d/go-kallax/pull/208/files). For some reason, however, Postgres runs this query much slower than if it were just doing a `count(*)`. I'm even able to replicate this behavior using the `pets` schema used in kallax's tests/benchmarks.

```text
testing=# select count(*) from pets;
 count
--------
 204957
(1 row)

testing=# explain analyze SELECT COUNT(*) FROM pets __pet WHERE __pet.kind = 'cat';
                                                                QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=3913.86..3913.87 rows=1 width=8) (actual time=22.457..22.457 rows=1 loops=1)
   ->  Gather  (cost=3913.75..3913.86 rows=1 width=8) (actual time=22.384..22.454 rows=2 loops=1)
         Workers Planned: 1
         Workers Launched: 1
         ->  Partial Aggregate  (cost=2913.75..2913.76 rows=1 width=8) (actual time=19.289..19.289 rows=1 loops=2)
               ->  Parallel Seq Scan on pets __pet  (cost=0.00..2813.66 rows=40035 width=0) (actual time=0.020..15.464 rows=34160 loops=2)
                     Filter: (kind = 'cat'::text)
                     Rows Removed by Filter: 68319
 Planning time: 0.077 ms
 Execution time: 24.007 ms
(10 rows)

testing=# explain analyze SELECT COUNT(__pet.*) FROM pets __pet WHERE __pet.kind = 'cat';
                                                                 QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=3913.86..3913.87 rows=1 width=8) (actual time=28.419..28.419 rows=1 loops=1)
   ->  Gather  (cost=3913.75..3913.86 rows=1 width=8) (actual time=28.352..28.416 rows=2 loops=1)
         Workers Planned: 1
         Workers Launched: 1
         ->  Partial Aggregate  (cost=2913.75..2913.76 rows=1 width=8) (actual time=24.702..24.702 rows=1 loops=2)
               ->  Parallel Seq Scan on pets __pet  (cost=0.00..2813.66 rows=40035 width=43) (actual time=0.059..20.927 rows=34160 loops=2)
                     Filter: (kind = 'cat'::text)
                     Rows Removed by Filter: 68319
 Planning time: 0.080 ms
 Execution time: 29.979 ms
(10 rows)

testing=#
```

There is about a 5ms difference with just over 200k rows in the `pets` table. We were able to consistently see examples where `count(__x.*)` is 2 - 5 seconds slower than `count(*)` in our production databases.

### DIFF
--- a/store.go
+++ b/store.go
@@ -439,13 +439,11 @@ func (s *Store) Reload(schema Schema, record Record) error {
 	return rs.Scan(record)
 }
 
-var all = NewSchemaField("*")
-
 // Count returns the number of rows selected by the given query.
 func (s *Store) Count(q Query) (count int64, err error) {
 	_, queryBuilder := q.compile()
 	builder := builder.Set(queryBuilder, "Columns", nil).(squirrel.SelectBuilder)
-	err = builder.Column(fmt.Sprintf("COUNT(%s)", all.QualifiedName(q.Schema()))).
+	err = builder.Column("COUNT(*)").
 		RunWith(s.runner).
 		QueryRow().
 		Scan(&count)


film42:
Postgres seems to handle `count(__alias.*)` different from `count(*)`. Take a look a the seq scan widths:
```
# Using an alias
->  Parallel Seq Scan on pets __pet  (cost=0.00..2813.66 rows=40035 width=43)
```

```
# Using the "*" constant
->  Parallel Seq Scan on pets __pet  (cost=0.00..2813.66 rows=40035 width=0)
```
The query planner is able to optimize `width=0` for `*` where it can't for `__alias.*`. On a bigger table with many rows, we've seen `width` peak around 600, which is what's causing the 2-5 sec slow-down over `*`.

Another factor is that the query planner has difficult preferring an index scan over a sequential scan when we use a table alias. With `COUNT(*)` we get a very fast index scan every time. With `COUNT(__alias.*)` it always picks a sequential scan. We've been scratching our brains as to why such a small change can have a meaningful impact in the query planner, which can be a magical black box, but based on the testing we've been doing, `COUNT(*)` wins every time.

film42:
Also, I believe `count(*)` will continue to work the way the previous change intended with a `left outer join`. The only difference where this won't produce the same result is with `right outer join` if the left-hand table's alias was previously referenced int he `count(__alias.*)` clause. This is because `NULL` values are excluded from count results.

So given two tables with no matching foreign keys:
```
# LEFT OUTER JOIN

# select count(t.*) from transactions t left join accounts a on t.account_id = a.id;
 count 
-------
     1
(1 row)

# select count(*) from transactions t left join accounts a on t.account_id = a.id;
 count 
-------
     1
(1 row)
```

```
# RIGHT OUTER JOIN

# select count(t.*) from transactions t right join accounts a on t.account_id = a.id;
 count 
-------
     0
(1 row)

# select count(*) from transactions t right outer join accounts a on t.account_id = a.id;
 count 
-------
   106
(1 row)
```

But I think this is OK because I don't think it's possible to build a right join in Kallax at the moment. So this change is backwards compatible as far as I can tell.

roobre:
Hello, 

First of all, let me thank @minond for tracing this and @film42 for the great explanation and feedback. As this seems to be backwards-compatible, I'm ok with merging it. It may take some time since a new release is made tho, since there are a few changes which require manual testing and right now I don't have much time for making them.

minond:
Thanks @roobre, we appreciate this library and all of the work that you put it to make it great.

PR #299, opened 2018-11-10
Fixes #298

### LINKED ISSUE #298: []byte fields get overwritten with random data when updating a model
Version: v1.3.5

[See fix below](https://github.com/src-d/go-kallax/issues/298#issuecomment-435564130) 

I have cooked up a [minimal test case](https://github.com/hayeah/go-kallax-bug). 

The test will increment `Counter` by 1, and we expect that `Data` and `Data2` would remain the same after the update.

```go
type TestModel struct {
	kallax.Model `table:"test_models"`
	ID           int64 `pk:"autoincr"`
	Data         []byte
	Data2        []byte
	Counter      int64
}
```

The `Update` method is called, limiting the change to `Counter`:

```
_, err = store.Update(obj, Schema.TestModel.Counter)
```

## Unexpected Output

Before updating the counter, fill object with random bytes:

```
obj.Data: abff8e1e90dd27df343ebef523e241326bd6a1490f960d07ee3095df7f6850e0
obj.Data2: c83aa6f5aa1f489df8ccaa55eac71eb293d27637be03a55d777fefcd24796f43
```

After the update, these should be the same, but is now changed to:

```
after update obj.Data: 0001310000000135000045000000090000000000530000000431205748455245
after update obj.Data2: 2432000000440000000753330053000000047637be03a55d777fefcd24796f43
```

## Analysis

This happens only if the object is fetched from the DB using a query:

```
obj = store.MustFindOne(NewTestModelQuery().FindByID(obj.ID))
```

This DOES NOT happen if the object is created from constructor:

```
obj := NewTestModel()
```

In fact, if [line#56](https://github.com/hayeah/go-kallax-bug/blob/1027d1a56f4f198a28309513833fa02bc97d3997/model.go#L56) from the test case is removed, it works as expected.

hayeah:


roobre:
Looks good to me, but I don't fully know the type internals so perhaps @erizocosmico wants to take a look at it too.

POD666:
Nice fix! It helped me a lot!
I had just looked throw `slices.go` and it looks like everywhere else `src` used to extract values, not just pointing to the same pointer.
So I suggest to merge it but mind that I'm new to Go and may be wrong.

hayeah:
@erizocosmico 

According to [sql.Scanner](https://golang.org/pkg/database/sql/#Scanner), the only reference types are `[]byte` and `time.Time`?

Not 100% sure about time.Time

```
// Reference types such as []byte are only valid until the next call to Scan
// and should not be retained. Their underlying memory is owned by the driver.
// If retention is necessary, copy their values before the next call to Scan.
```

